PR #1, opened 2022-07-29
Flagd provider 

james-milligan:
https://github.com/open-feature/golang-sdk-contrib/issues/2


AlexsJones:
Can I get added as reviewer ?

providers/flagd/pkg/provider.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+}
+
+func WithHost(host string) ProviderOption {
+	return func(p *Provider) {
+		p.host = host
+	}
+}
+
+// service should be one of "http" or "grpc", if not the default "http" will be used
```
warber, line 57:
```suggestion
// WithService specifies the type of the service. service should be one of "http" or "grpc", if not the default "http" will be used
```

to make the linters happy, comments of exported functions shall start with their name.
Alos, consider introducing kind of an enum `ServiceType` to improve ux, e.g:
```
type ServiceType string

const (
	HTTP ServiceType = "http"
	GRPC             = "grpc"
)
```

providers/flagd/pkg/provider.go
```diff
@@ -0,0 +1,161 @@
...
+func WithService(service string) ProviderOption {
+	return func(p *Provider) {
+		p.serviceName = service
+	}
+}
+
+// service should be one of "http" or "https", if not the default "http" will be used, https is not currently supported
```
warber, line 64: same comments as in line 57

providers/flagd/pkg/service/grpc/constructor.go
```diff
@@ -0,0 +1,34 @@
+package grpc_service
+
+type GRPCServiceOption func(*GRPCServiceConfiguration)
+
+func NewGRPCService(opts ...GRPCServiceOption) *GRPCService {
+	const (
+		port = 8080
+		host = "localhost"
```
warber, line 8: `host` is not used

providers/flagd/README.md
```diff
@@ -0,0 +1,41 @@
...
+        flagd.WithHost("localhost"),
+        flagd.WithPort(8000),
+    ))
+}
+
+```
+
+## License
+
+Apache 2.0 - See [LICENSE](./license) for more information.
```
warber, line 41: this file does not exist

providers/flagd/pkg/service/grpc/constructor.go
james-milligan: updated

providers/flagd/README.md
james-milligan: updated path

providers/flagd/pkg/service/grpc/client.go
```diff
@@ -0,0 +1,42 @@
...
+		)
+		if err != nil {
+			log.Errorf("grpc - fail to dial: %v", err)
+			return
+		}
+		s.conn = conn
+	}
+}
+
+func (s *GRPCClient) GetInstance() schemaV1.ServiceClient {
```
skyerus, line 36: Would prefer `Instance()` as the name here

providers/flagd/pkg/service/grpc/service.go
```diff
@@ -0,0 +1,175 @@
+package grpc_service
+
+import (
+	ctx "context"
+	"errors"
+
+	models "github.com/open-feature/flagd/pkg/model"
+	models2 "github.com/open-feature/golang-sdk-contrib/providers/flagd/pkg/model"
```
skyerus, line 8: Can we think of any better naming than `models`/`models2`?

providers/flagd/pkg/service/grpc/service.go
```diff
@@ -0,0 +1,175 @@
...
+	Port int32
+	Host string
+}
+
+type GRPCService struct {
+	Client IGRPCClient
+}
+
+func (s *GRPCService) ResolveBoolean(flagKey string, context of.EvaluationContext, options ...service.IServiceOption) (*schemaV1.ResolveBooleanResponse, error) {
+	client := s.Client.GetInstance()
```
skyerus, line 27: There's a fair amount of repeated code in these funcs, can this be simplified?

providers/flagd/pkg/service/grpc/tests/service-resolve-boolean_test.go
```diff
@@ -0,0 +1,154 @@
+package tests
```
skyerus, line 1: What's the reason for moving tests into their own directory? I've not seen this approach before, [generally tests are included in the same dir](https://go.dev/doc/tutorial/add-a-test) as what they're testing with the same package name appended with `_test`. It'd be good to remain consistent

providers/flagd/pkg/service/grpc/tests/mocks/mock-service-client.go
```diff
@@ -0,0 +1,138 @@
+package mocks
```
skyerus, line 1: Why not use https://github.com/golang/mock for this?

providers/flagd/pkg/service/http/client.go
```diff
@@ -0,0 +1,30 @@
+package http_service
+
+import (
+	"io"
+	"net/http"
+)
+
+type IHTTPClient interface {
```
skyerus, line 8: ought to have a comment as its exported (same for the rest)

providers/flagd/pkg/provider.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+}
+
+func (p *Provider) Metadata() of.Metadata {
+	return of.Metadata{
+		Name: "flagd",
+	}
+}
+
+func (p *Provider) GetBooleanEvaluation(flagKey string, defaultValue bool, evalCtx of.EvaluationContext, options ...of.EvaluationOption) of.BoolResolutionDetail {
```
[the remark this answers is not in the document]
james-milligan: updated to latest go-sdk version and signatures updated

providers/flagd/pkg/service/grpc/service.go
james-milligan: renamed to (package)Models

providers/flagd/pkg/service/grpc/tests/service-resolve-boolean_test.go
james-milligan: updated to have tests in the same directory as the corresponding services

providers/flagd/pkg/service/http/client.go
james-milligan: updated exported / unexported structs / methods following the test changes, and added comments to exported values

providers/flagd/pkg/provider.go
```diff
@@ -0,0 +1,170 @@
...
+
+// WithService specifies the type of the service. service should be one of "http" or "grpc", if not the default "http" will be used
+func WithService(service ServiceType) ProviderOption {
+	return func(p *Provider) {
+		p.serviceName = service
+	}
+}
+
+// WithProtocol specifies the protocol used by the http service. Should be one of "http" or "https", if not the default "http" will be used, https is not currently supported
+func WithProtocol(protocol string) ProviderOption {
```
warber, line 74: nitpick: I think you could introduce a `Protocol` enum here, like you did for `ServiceType`. Please also adapt the comments speaking about what can be passed as a value.... (same for WithService(...))

james-milligan: updated in latest commit

providers/flagd/pkg/service/grpc/service.go
james-milligan: I believe that without access to generics trying to simplify this code would be difficult, the response type needs to match that of the http service and as a result the response structures would need to be constructed outside of the helpers - which would leave the code in a reasonably similar position to how it is at the moment.
Maybe this can still be done once the code base is moved to 1.18

providers/flagd/pkg/service/grpc/tests/service-resolve-boolean_test.go
warber: imo this still looks a bit unfamiliar. Since you are basically just testing the logic/methods of the `HTTP/GRPCService` that are living inside the `service.go` files,  I would propose to move all the tests to a corresponding `service_test.go` file.

providers/flagd/pkg/service/http/resolve-number_test.go
```diff
@@ -0,0 +1,154 @@
+package http_service
```
skyerus, line 1: These ought to be in a test package: `package http_service_test`

providers/flagd/pkg/provider.go
warber: Mhm...., I agree that it doesn't look nice. Looking at it, does it make sense to cover all of this by just providing a `WithProtocol` option ? :thinking: 
```
NewProvider(WithProtocol(flagd.HTTP))) or
NewProvider(WithProtocol(flagd.HTTPS))) or
NewProvider(WithProtocol(flagd.GRPC))
```

james-milligan: could also update the enum to be evaluated at the top level?

```
type Provider struct {
	service     service.IService
	port        int32
	host        string
	serviceName ServiceType
	protocol    string
}

func WithProtocol(protocol HTTPService.ProtocolType) ProviderOption {
	if protocol == HTTPS {
		return func(p *Provider) {
			p.protocol = "https"
		}
	}
	return function(p *Provider) {
		p.protocol = "http"
	}
}
```

so the HTTP service is passed the correct string value and is unaware of the provider level enum

providers/flagd/pkg/service/grpc/tests/service-resolve-boolean_test.go
james-milligan: updated all service tests to sit within the `service_test.go` file(s)

providers/flagd/pkg/service/grpc/tests/mocks/mock-service-client.go
james-milligan: updated tests to use https://github.com/golang/mock

providers/flagd/pkg/service/http/resolve-number_test.go
james-milligan: updated

providers/flagd/pkg/provider.go
warber: hm yeah i like the idea :) However, one thing which bugs me is that, `protocol` is a field of `Provider` whereas it is actually only applicable to HTTP based service implementation. (correct?). Then, for me it would indeed make sense to just remove this field `protocol` from the Provider level and distinguish between three different types of services e.g. using `NewProvider(WithService(HTTP|HTTPS|GRPC))` on the provider level, and get rid of `WithProtocol`. Or do you have different view on this? Maybe additional oppinions would help. @skyerus ? :-)

james-milligan: that's a good point, I think it will also be the cleanest implementation, you've convinced me!

providers/flagd/pkg/service/http/constructor.go
```diff
@@ -0,0 +1,57 @@
...
+	HTTP ProtocolType = iota
+	HTTPS
+)
+
+// NewHTTPService creates a new HTTPService taking configuration options to overide default values
+func NewHTTPService(opts ...HTTPServiceOption) *HTTPService {
+	const (
+		port     = 8080
+		host     = "localhost"
+		protocol = HTTP
```
warber, line 17: unused constant

providers/flagd/pkg/provider.go
```diff
@@ -0,0 +1,182 @@
...
+	} else {
+		provider.Service = HTTPService.NewHTTPService(
+			HTTPService.WithPort(provider.providerConfiguration.Port),
+			HTTPService.WithHost(provider.providerConfiguration.Host),
+		)
+	}
+	return provider
+}
+
+// WithHost specifies the port of the flagd server. Defaults to 8080
```
warber, line 65: pls fix comment

providers/flagd/pkg/service/http/service.go
```diff
@@ -0,0 +1,133 @@
...
+	err := s.FetchFlag(url, context, &resMess)
+	if err != nil {
+		return &schemaV1.ResolveBooleanResponse{
+			Reason: models.ErrorReason,
+		}, err
+	}
+	return &resMess, nil
+}
+
+// ResolveBoolean handles the flag evaluation response from the grpc flagd flags/{flagKey}/resolve/string endpoint
```
warber, line 41: pls fix comment(s)

providers/flagd/pkg/service/grpc/service.go
```diff
@@ -0,0 +1,179 @@
...
+	providerModels "github.com/open-feature/golang-sdk-contrib/providers/flagd/pkg/model"
+	of "github.com/open-feature/golang-sdk/pkg/openfeature"
+	log "github.com/sirupsen/logrus"
+	schemaV1 "go.buf.build/grpc/go/open-feature/flagd/schema/v1"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+type GRPCServiceConfiguration struct {
+	Port int32
```
warber, line 17: same here, i think `uint16` makes more sense. If you agree pls also check other places. :)

providers/flagd/pkg/provider.go
```diff
@@ -8,18 +8,23 @@ import (
...
-	port        int32
-	host        string
-	serviceName ServiceType
-	protocol    HTTPService.ProtocolType
+	Service               service.IService
+	providerConfiguration *ProviderConfiguration
+}
+
+type ProviderConfiguration struct {
+	Port        int32
```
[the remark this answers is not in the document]
james-milligan: updated in all cases

providers/flagd/pkg/service/http/constructor.go
james-milligan: removed

providers/flagd/pkg/provider.go
james-milligan: fixed

providers/flagd/pkg/service/http/service.go
james-milligan: fixed

providers/flagd/pkg/service/grpc/service.go
james-milligan: updated :)

providers/flagd/pkg/service/http/service.go
```diff
@@ -0,0 +1,135 @@
...
+	return errors.New(models.GeneralErrorCode)
+}
+
+func flattenContext(ctx of.EvaluationContext) map[string]interface{} {
+	if ctx.TargetingKey != "" {
+		ctx.Attributes["TargettingKey"] = ctx.TargetingKey
+	}
+	if ctx.Attributes == nil {
+		return map[string]interface{}{}
+	}
```
warber, line 133: if ctx.Attributes is indeed `nil` the call at line `129` would produce a panic.
You could just exchange the order of the two if statement (checking for nil of ctx.Attributes frst)

james-milligan: You're right, the reason I wrote it out in this way was in the case of the `EvaluationContext` definition growing and the `if ctx.Attributes == nil` needing to be evaluated multiple times, however at the moment there isn't any need for this and can refactor back when/if the definition is expanded

warber: Sorry, I think i created a bit of confusion, what I meant is something like this:
```
func flattenContext(ctx of.EvaluationContext) map[string]interface{} {
	if ctx.Attributes == nil {
		return map[string]interface{}{}
	}

	if ctx.TargetingKey != "" {
		ctx.Attributes["TargetingKey"] = ctx.TargetingKey
	}

	return ctx.Attributes
}
```
it's a bit of a nitpick but I think that's the cleanest logic:  if `ctx.Attributes` is `nil`, it makes no sense to search for a specific key in it. So you can just return early from this func :)

james-milligan: oh sorry, the purpose of this function is to return a single `map[string]interface{}` from the `EvaluationContext` struct. The struct definition defines `TargetingKey` in order to ensure that it remains constant across vendors. In the instance that `ctx.Attributes` is nil but the `TargetingKey` is not, then the returned `map[string]interface{}` must still contain this field.
```
EvaluationContext{
    TargetingKey: "James",
    Attributes: nil,
}
```
should give
```
map[string]interface{}{
    "TargetingKey": "james",
}
```
Its a bit of an odd situation and I've raised an issue regarding it [here](https://github.com/open-feature/golang-sdk/issues/30#issue-1322347254)

warber: ah understood, now also the func name makes more sense to me :D  thx :)

providers/flagd/pkg/service/http/service.go
```diff
@@ -0,0 +1,135 @@
...
+	if err != nil {
+		return &schemaV1.ResolveObjectResponse{
+			Reason: models.ErrorReason,
+		}, err
+	}
+	return &resMess, nil
+}
+
+// FlagKey handles the calling and parsing of the flag evaluation endpoints.
+// Argument p should be a pointer to a valid Resolve{type}Response struct for unmarshalling the response, e.g ResolveObjectResponse{}.
```
warber, line 81: pls fix comment

providers/flagd/pkg/service/grpc/client_mock_test.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+func (m *MockClient) Instance() schemaV1.ServiceClient {
+	if m.NilClient {
+		return nil
+	}
+	return m.Client
+}
+
+func (s *MockClient) Configuration() *service.GRPCServiceConfiguration {
```
warber, line 20: pls align receiver name to `m`

providers/flagd/pkg/provider.go
```diff
@@ -0,0 +1,182 @@
...
+	Host        string
+	ServiceName ServiceType
+}
+
+type ServiceType int
+
+const (
+	// HTTP argument for use in WithService, this is the default value
+	HTTP ServiceType = iota
+	// HTTPS argument for use in WithService, overides the default value of http (NOT IMPLEMENTED BY FLAGD)
```
warber, line 26: typo "override" , pls also fix other occurences

providers/flagd/pkg/service/grpc/constructor_test.go
```diff
@@ -0,0 +1,71 @@
...
+			protocol: "http",
+			options: []service.GRPCServiceOption{
+				service.WithPort(1),
+			},
+		},
+	}
+	for _, test := range tests {
+		svc := service.NewGRPCService(test.options...)
+		if svc == nil {
+			t.Error("recieved nil service from NewGRPCService")
```
warber, line 48: typo received, pls fix also other occurences

warber (commented):
final round - mainly some typos :)

warber (approved):
looks good to me. nice work :+1:

skyerus (approved):
LGTM - nice work!

beeme1mr (approved):

PR #15, opened 2022-08-25
Validator hook

skyerus:
With hex color validator


hooks/validator/pkg/validator/validator.go
```diff
@@ -0,0 +1,31 @@
...
+type validator interface {
+	IsValid(flagEvaluationDetails of.EvaluationDetails) error
+}
+
+// Hook validates the flag evaluation details After flag resolution
+type Hook struct {
+	Validator validator
+}
+
+func (h Hook) Before(hookContext of.HookContext, hookHints of.HookHints) (*of.EvaluationContext, error) {
```
beeme1mr, line 16: Is it required to add unused hooks?

james-milligan: The methods will still need to be present to satisfy the interface even if they are unused

beeme1mr: ok, thanks.

james-milligan (approved):
nice work :)

hooks/validator/README.md
```diff
@@ -0,0 +1,86 @@
...
+func main() {
+	v := validator.Hook{Validator: hex.Validator{}}
+}
+```
+
+Use the validator hook (on invocation as an example):
+```go
+client := openfeature.NewClient("foo")
+evalOptions := openfeature.NewEvaluationOptions([]openfeature.Hook{v}, openfeature.HookHints{})
+value, err := client.
```
beeme1mr, line 39: Can you also confirm that value contains the fallback string `#0000FF`?

skyerus: I'm assuming you mean in the case where the resolved flag is invalid?
Otherwise that's the default behaviour

hooks/validator/README.md
```diff
@@ -64,18 +64,21 @@ func main() {
...
+	fmt.Println("result:", result)
 }
 ```
 
 ```shell
 go run main.go
 err: execute after hook: invalid format
+result {blue 1 {invalidhex   }}
 ```
+Note that despite getting an error we still get a result.
```
beeme1mr, line 81: Nice! Thanks for explicitly calling this out. It may not be obvious to newcomers but it's important to understand.

hooks/validator/README.md
```diff
@@ -0,0 +1,89 @@
+# Validator Hook
+
+![Experimental](https://img.shields.io/badge/experimental-breaking%20changes%20allowed-yellow)
+![Alpha](https://img.shields.io/badge/alpha-release-red)
```
beeme1mr, line 4: It's up to you but these badges can likely be removed.

skyerus: Done

hooks/validator/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+The `validator hook` taps into the `After` lifecycle event to validate the result of flag evaluations.
+
+The hook defines a `validator` interface with function
+```go
+IsValid(flagEvaluationDetails of.EvaluationDetails) error
+```
+to allow application authors to supply their own validators.
+
+There are, however, [ready to be used validators](#validators) that conform to the interface.
```
toddbaert, line 11: An optional function with ready-to-use validators is a great way to go, nice work.

hooks/validator/README.md
```diff
@@ -0,0 +1,86 @@
+# Validator Hook
+
+The `validator hook` taps into the `After` lifecycle event to validate the result of flag evaluations.
```
toddbaert, line 3: @skyerus @beeme1mr think it's worth mentioning right here that it defaults if it's invalid?

Maybe that's obvious and takes away from this otherwise very snappy sentence. :shrug:

hooks/validator/pkg/hex/hex.go
```diff
@@ -0,0 +1,42 @@
...
+	return nil
+}
+
+func validateByte(b byte) error {
+	if (b >= '0' && b <= '9') || (b >= 'a' && b <= 'f') || (b >= 'A' && b <= 'F') {
+		return nil
+	}
+
+	return errInvalidFormat
+}
```
toddbaert, line 42: Another way to implement this could be to create a regex validator, and then supply a regex to that that matches CSS hex values.

toddbaert (approved):

hooks/validator/README.md
skyerus: Done!

hooks/validator/pkg/hex/hex.go
skyerus: Great suggestion, implemented :)

beeme1mr (approved):

PR #19, opened 2022-08-26
Flagd spec v0.4.0 compliance + updated go-sdk version

james-milligan:


providers/flagd/pkg/provider.go
```diff
@@ -108,6 +103,11 @@ func WithService(service ServiceType) ProviderOption {
 	}
 }
 
+// Hooks flagd provider does not have any hooks, returns empty slice
```
skyerus, line 106: curious how this ended up moving

skyerus (approved):

beeme1mr (approved):

PR #28, opened 2022-10-05
feat!: introduced connect protocol for flagd provider

james-milligan:
## This PR
- flagd provider now uses the connect protocol


providers/flagd/pkg/service/service.go
```diff
@@ -0,0 +1,180 @@
...
+		}, of.NewParseErrorResolutionError(err.Error())
+	}
+	res, err := client.ResolveObject(ctx, connect.NewRequest(&schemaV1.ResolveObjectRequest{
+		FlagKey: flagKey,
+		Context: evalCtxF,
+	}))
+	if err != nil {
+		return &schemaV1.ResolveObjectResponse{
+			Reason: string(of.ErrorReason),
+		}, errors.New(handleError(err))
```
skyerus, line 161: is it possible to return these errors as an openfeature.ResolutionError so they can be picked up on at the outer level?

james-milligan: done :)

skyerus (approved):
love to see all that code deletion

beeme1mr (approved):
Looks good to me! I updated the PR to specify that this is a breaking change.

providers/flagd/pkg/service/service.go
```diff
@@ -0,0 +1,179 @@
...
+	schemaV1 "go.buf.build/open-feature/flagd-connect/open-feature/flagd/schema/v1"
+	"google.golang.org/protobuf/types/known/structpb"
+)
+
+type ServiceConfiguration struct {
+	Port            uint16
+	Host            string
+	CertificatePath string
+	SocketPath      string
+}
```
toddbaert, line 21: Do we need a boolean for `Tls`?

I would assume that if the remote flagd has a cert signed by a popular CA, default system certs would be used, so there'd be no need for a `CertificatePath`. In that case, if you want to use TLS, what do you do? Or am I wrong and you always need to provide certs? That would surprise me.

toddbaert: see https://github.com/open-feature/js-sdk-contrib/blob/321ac0936c6744c6cf9956c6708a701a3d9237be/libs/providers/flagd/src/lib/configuration.spec.ts#L25-L28

james-milligan:
https://github.com/open-feature/go-sdk-contrib/issues/31

PR #48, opened 2022-11-30
feat: implement caching in flagd provider

### LINKED ISSUE #44: Implement basic caching strategy for flagd provider
Follow the strategy defined in https://github.com/open-feature/flagd/issues/223

skyerus:
## This PR

Implements caching of the flagd provider as defined by [this strategy](https://github.com/open-feature/flagd/blob/main/docs/caching.md).

### Related Issues

Fixes #44


providers/flagd/pkg/cache/in_memory.go
```diff
@@ -0,0 +1,50 @@
+package cache
+
+import (
+	"context"
+	"sync"
+)
+
+type InMemory struct {
```
skyerus, line 8: This is a simple in memory implementation, good for testing purposes and can be used in cases where growing cache isn't a concern. In cases where cache size is a concern I'd suggest using a least recently used (LRU) caching implementation like [hashicorp's golang-lru](https://github.com/hashicorp/golang-lru).
Any thoughts against introducing this package in the next phase? @toddbaert @james-milligan @beeme1mr @AlexsJones

providers/flagd/pkg/provider.go
```diff
@@ -118,6 +132,27 @@ func WithHost(host string) ProviderOption {
...
+
+// WithCustomCache applies a custom cache implementation
+func WithCustomCache(cacheImplementation Cache) ProviderOption {
+	return func(p *Provider) {
+		p.Cache = cacheImplementation
+	}
+}
+
+// WithContext supplies the given context to the event stream
+func WithContext(ctx context.Context) ProviderOption {
```
skyerus, line 150: I've opted to allow context to be included via an option rather than the idiomatic first function parameter for the sake of not being a breaking change.

james-milligan: I like this approach

providers/flagd/pkg/cache/in_memory.go
toddbaert: 1.19 brings [sync.Map](https://pkg.go.dev/sync#Map) , but I don't know if we want to upgrade just for this.

providers/flagd/pkg/cache/in_memory.go
```diff
@@ -0,0 +1,50 @@
...
+	values map[string]interface{}
+	rwMux  *sync.RWMutex
```
toddbaert, line 10: 1.19 brings [sync.Map](https://pkg.go.dev/sync#Map) , but I don't know if we want to upgrade just for this.

providers/flagd/pkg/constant/reason.go
```diff
@@ -0,0 +1,5 @@
+package constant
+
+const (
+	ReasonCached = "CACHED"
```
toddbaert, line 4: Do you think it's worth adding this `reason` to the spec? Or just to the SDKs?

I'm not sure it's worth the effort, but I'm interested in your opinion.

providers/flagd/pkg/cache/in_memory.go
skyerus: I did consider it initially but
> The Map type is optimized for two common use cases: (1) when the entry for a given key is only ever written once but read many times, as in caches that only grow, or (2) when multiple goroutines read, write, and overwrite entries for disjoint sets of keys. In these two cases, use of a Map may significantly reduce lock contention compared to a Go map paired with a separate Mutex or RWMutex.

convinced me that it wasn't the right use case.
Either way I've since updated this to use generics, will push up soon when I finish the tests 😄

toddbaert: > Either way I've since updated this to use generics

Nice! I think I went through the same thing with the flagd-js provider, as I added caching, etc, It became necessary in order to avoid too much duplication.

toddbaert:
@skyerus do you think we should also enable setting the cache from a standardized environment variable, like some of the other settings?>

providers/flagd/pkg/cache/in_memory.go
toddbaert: I guess another candidate might be https://github.com/orcaman/concurrent-map

providers/flagd/pkg/constant/reason.go
beeme1mr: Yeah, I think it makes sense.

skyerus: +1 from me, I think it makes sense to define the behaviour to avoid discrepancies between implementations.

providers/flagd/pkg/provider.go
```diff
@@ -271,11 +429,81 @@ func (p *Provider) ObjectEvaluation(
...
+	if !ok {
+		return errors.New("no flagKey field in event data")
+	}
+
+	flagKey, ok := flagKeyVal.(string)
+	if !ok {
+		return errors.New("flagKey is not a string")
+	}
+
+	// TODO: consider sending the flag type in the configuration change event
```
skyerus, line 501: potential optimisation

providers/flagd/pkg/constant/reason.go
skyerus: Created an issue for this: https://github.com/open-feature/spec/issues/165

james-milligan (approved):

providers/flagd/README.md
```diff
@@ -16,6 +16,9 @@ flagd.WithHost(string)                  // defaults to localhost
 flagd.WithPort(uint16)                  // defaults to 8013
 flagd.FromEnv()                         // sets the provider configuration from environment variables
 flagd.WithSocketPath(string)            // no default, when set a unix socket connection is used (only available for GRPC)
+flagd.WithoutCache()                    // disables caching of flag evaluations
+flagd.WithLRUCache(1000)                // enables LRU caching (see configuring caching section)
+flagd.WithBasicInMemoryCache()          // enables basic in memory cache (see configuring caching section)
```
james-milligan, line 21: might it be worthwhile to add a footnote here to mention that the latest option will take precedence, although as previously mentioned this may be overkill, e.g. 
`flagd.NewProvider(flagd.WithoutCache(), flagd.WithLRUCache(1000))`

providers/flagd/README.md
```diff
@@ -35,12 +38,49 @@ func main() {
 ```
 
 ### Using flagd.FromEnv()  
-By default the flagd provider will not read environment variables to set its own configuration, however, if the `flagd.FromEnv()` option is set as an argument for the `flagd.NewProvider()` method, then the following environment variables will be checked: `FLAGD_HOST`, `FLAGD_PORT`, `FLAGD_SERVER_CERT_PATH`.
+By default the flagd provider will not read environment variables to set its own configuration, however, if the `flagd.FromEnv()` option is set as an argument for the `flagd.NewProvider()` method, then the following environment variables will be checked: `FLAGD_HOST`, `FLAGD_PORT`, `FLAGD_SERVER_CERT_PATH` & `FLAGD_CACHING_DISABLED`.
```
toddbaert, line 41: I think enabling caching by default is a safe option with this primitive form of caching (caching only static resolutions). One thing to consider is future caching enhancements that we might not want on by default, or wherein we'd need more than a simple "on/off" config... it's hard to change the default later.

providers/flagd/pkg/provider.go
```diff
@@ -118,6 +139,79 @@ func WithHost(host string) ProviderOption {
...
+		p.stringCache = cache.NewInMemory[string, of.StringResolutionDetail]()
+		p.floatCache = cache.NewInMemory[string, of.FloatResolutionDetail]()
+		p.interfaceCache = cache.NewInMemory[string, of.InterfaceResolutionDetail]()
+
+		p.cacheEnabled = true
+	}
+}
+
+// WithLRUCache applies least recently used caching (github.com/hashicorp/golang-lru).
+// The provided size is the limit of the number of cached values for each type of flag. Once the limit is reached each
```
toddbaert, line 168: I really thought "size" was "size in bytes", but seems it's actually "number of entries". The LRU cache doc isn't very clear about this.

providers/flagd/pkg/provider.go
```diff
@@ -271,11 +429,81 @@ func (p *Provider) ObjectEvaluation(
...
+					log.Errorf("handle configuration change event: %v", err)
+				}
+			}
+		case err := <-errChan:
+			if p.cacheEnabled { // disable cache
+				WithoutCache()(p)
+			}
+			return err
+		}
+	}
```
toddbaert, line 479: What happens here in the case of a disconnection/re-connection? At least in the Java impl, the underlying gRPC libraries nicely handle re-connection. I assume that's the case here, but is it possible that we miss a flag change event while disconnected?

I think we may want to enhance flagd to send the latest flag config on connection, but a minimal solution might be to `Purge` the cache on reconnect. What do you think?

providers/flagd/README.md
skyerus: Imo this is overkill, I can't imagine a scenario where a user would supply both options

providers/flagd/README.md
skyerus: Are you saying you think it's safer to disable by default in anticipation of potential future enhancements?

toddbaert: I was basically thinking out loud about that question. I believe enabling it by default is still probably the right choice, I just wanted to point out that it could make future enhancements a bit more difficult.

providers/flagd/pkg/service/service.go
```diff
@@ -190,21 +197,46 @@ func handleError(err error) openfeature.ResolutionError {
...
+			delay = 2 * delay
+			log.Infof("connection to event stream failed, sleeping %v", delay)
+			time.Sleep(delay)
+		}
+	}
+
+	if err != nil {
+		errChan <- err
+	}
+}
```
toddbaert, line 221: I wonder if messages related to reconnect and giving up connection should be `warning` level.

toddbaert (approved):

providers/flagd/pkg/service/service.go
```diff
@@ -190,21 +197,46 @@ func handleError(err error) openfeature.ResolutionError {
...
+	delay := s.baseRetryDelay
+	var err error
+	for i := 1; i <= maxAttempts; i++ {
+		log.Infof("attempt %d at connecting to event stream", i)
+		i, err = s.eventStream(ctx, eventChan, i, maxAttempts)
+		if i == 1 {
+			delay = s.baseRetryDelay // reset delay if the connection was successful before failing
+		}
+		if err != nil && i <= maxAttempts {
+			delay = 2 * delay
```
james-milligan, line 212: some form of offset should be used here, especially if a future deployment model is n:1

toddbaert: Ins't that what `delay = 2 * delay` is? This would be a exponential (order 2) back-off no?

providers/flagd/pkg/provider.go
skyerus: Great shout, I've now implemented this as you suggested with a retry mechanism.

providers/flagd/pkg/service/service.go
james-milligan: As far as I'm aware its best practice to have an exponential back off with some degree of randomisation, otherwise there is potential for all deployments to be in sync, so if the reason for the failure is throttling then there is no chance for recovery. (this may be overkill of course)

providers/flagd/README.md
skyerus: I'm conflicted on this too. Anyone else have an opinion? @james-milligan @beeme1mr @AlexsJones @Kavindu-Dodan

skyerus:
Draft until https://github.com/open-feature/flagd/pull/229 has been merged and integrated here

providers/flagd/README.md
james-milligan: my vote is disabled by default, if we are worried about introducing debt its better to play it safe

beeme1mr: I think it should be enabled by default. We'll need to consider caching for any future enhancements anyways.

providers/flagd/pkg/provider.go
toddbaert: This is impacted by https://github.com/open-feature/flagd/issues/238

Issue #49, opened 2022-12-02
[otel-hook] Update OTel Hook to conform to official conventions


toddbaert:
Now that [OTel 2529](https://github.com/open-telemetry/opentelemetry-specification/pull/2529) is merged, we should update the Open Telemetry hook (currently exists, but not consistent with the convention).

Basically, this would entail updating the hook to:

- in the `after` stage: gets the current span and adds an appropriately named event (`feature_flag`), with the attributes (`feature_flag.key`, `feature_flag.provider_name`, and `feature_flag.variant`)
- in the `error` stage, records the exception

In pseudocode, this looks something like:

```
class OTelHook implements Hook {
  after(context, details) {
    var currentTrace = trace.getActiveSpan();
    currentTrace.addEvent("feature_flag", { "feature_flag.key": details.key, /*  other props */ } );
  }

  error(err) {
    var currentTrace = trace.getActiveSpan();
    currentTrace.recordException(err);
  }
}
```
See: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/semantic_conventions/feature-flags.md

PR #87, opened 2023-01-09
feat!: Update OTel Hook to conform to official conventions

james-milligan:
Signed-off-by: James Milligan <james@omnant.co.uk>

## This PR

- Updates OTel Hook to conform to official conventions

### Related Issues

https://github.com/open-feature/go-sdk-contrib/issues/49


hooks/open-telemetry/pkg/otel.go
```diff
@@ -2,165 +2,60 @@ package otel
...
-}
-
-// WithContext sets the parent context used when new spans are formed. When the parent context is closed so are all spans.
-// if no parent context is provided context.Background() is used internally
-func (h *Hook) WithContext(ctx context.Context) {
-	h.ctx = ctx
-}
-
-// Before creates the flag evaluations open-telemetry span and sets the FlagKey and ProviderName attributes
+// Before method is unused for this hook
```
skyerus, line 29: Could include `UnimplementedHook` from the go-sdk in your Hook struct to avoid implementing empty funcs

james-milligan: updated

hooks/open-telemetry/pkg/otel.go
```diff
@@ -2,165 +2,53 @@ package otel
...
-	AttributeFlagKey = "feature_flag.flag_key"
-	// AttributeProviderName trace attribute name for the provider used for evaluation
-	AttributeProviderName = "feature_flag.provider_name"
-	// AttributeEvaluatedVariant trace attribute name for the variant returned from the flag evaluation
-	AttributeEvaluatedVariant = "feature_flag.evaluated_variant"
-	// AttributeEvaluatedValue trace attribute name for the value returned from the flag evaluation
-	AttributeEvaluatedValue = "feature_flag.evaluated_value"
-	// AttributeTraceName the name of the trace exported by the hook
-	AttributeTraceName = "github.com/open-feature/go-sdk-contrib/hooks/open-telemetry/pkg"
+	EventName                 = "feature_flag"
```
skyerus, line 12: Do these need to be exported?

james-milligan: Ive kept this exported so that the test package can be kept separate, whilst checking for the correct value

skyerus (approved):

hooks/open-telemetry/pkg/otel.go
```diff
@@ -2,165 +2,53 @@ package otel
...
-	}
-	mw.ss.span.SetAttributes(
-		attribute.String(AttributeEvaluatedValue, value),
+	span := trace.SpanFromContext(h.ctx)
+	span.SetAttributes(
+		attribute.String(EventPropertyFlagKey, hookContext.FlagKey()),
+		attribute.String(EventPropertyProviderName, hookContext.ProviderMetadata().Name),
+		attribute.String(EventPropertyVariant, flagEvaluationDetails.Variant),
 	)
+	span.AddEvent(EventName)
```
beeme1mr, line 42: It looks like `AddEvent` accepts attributes. Here's an example: https://github.com/MrAlias/opentelemetry-go/blob/46582921368bcafa47ae4d5b17cca9b9e20e0a5a/example/basic/main.go#L91

hooks/open-telemetry/pkg/otel.go
```diff
@@ -2,165 +2,53 @@ package otel
...
-			return err
-		}
-		value = string(val)
-	default:
-		return fmt.Errorf("unknown data type received: %d", flagEvaluationDetails.FlagType)
-	}
-	mw.ss.span.SetAttributes(
-		attribute.String(AttributeEvaluatedValue, value),
+	span := trace.SpanFromContext(h.ctx)
+	span.SetAttributes(
```
beeme1mr, line 37: This doesn't seem correct. The attributes should be associated with the span event not the span.

beeme1mr (changes requested):
This doesn't appear to follow the OTel spec. The attributes must be on the event not the span itself.

james-milligan:
> This doesn't appear to follow the OTel spec. The attributes must be on the event not the span itself.

Ive made the requested changes, the attributes are now associated with the event, rather than the span itself

hooks/open-telemetry/pkg/otel.go
```diff
@@ -2,165 +2,52 @@ package otel
...
-		return errors.New("no span stored for provided hook context")
-	}
-	if flagEvaluationDetails.ResolutionDetail.Variant != "" {
-		mw.ss.span.SetAttributes(
-			attribute.String(AttributeEvaluatedVariant, flagEvaluationDetails.ResolutionDetail.Variant),
-		)
+// After sets the feature_flag event and associated attributes on the span stored in the context
+func (h *Hook) After(hookContext openfeature.HookContext, flagEvaluationDetails openfeature.InterfaceEvaluationDetails, hookHints openfeature.HookHints) error {
+	if h.ctx == nil {
+		// if no context is set trace.SpanFromContext will return a noop span
```
toddbaert, line 33: Is it better to simply return the no-op span then?

I'm trying to understand the benefits of handling this ourselves by returning `nul`, vs avoiding this check and returing a noop span.

hooks/open-telemetry/pkg/otel.go
```diff
@@ -2,165 +2,52 @@ package otel
...
-type storedSpan struct {
-	cancel func()
-	span   trace.Span
+	ctx context.Context
+	openfeature.UnimplementedHook
 }
 
 // NewHook return a reference to a new instance of the OpenTelemetry Hook
-func NewHook() *Hook {
+func NewHook(ctx context.Context) *Hook {
```
toddbaert, line 24: Should we somehow warn of an invalid state if this constructor is not used and the context is unset?

james-milligan: Unexpected the hook struct so the use of `NewHook` is forced, past this point I think the log is unnecessary, as you would have to pass a nil context into the constructor as you can no longer bypass this by creating the hook directly with `hook := otelhook.Hook{}`

hooks/open-telemetry/pkg/otel.go
james-milligan: ive removed the nil check and allowed for the loop hook to be used

beeme1mr (approved):
Approved by please fix the linting error.

PR #89, opened 2023-01-09
refactor: Handle context done channel

Kavindu-Dodan:
This was observed while attempting to introduce unit tests on flagD configuration change event handling. 

Further, removed an unused variable within `Service` event handler.

### DIFF
--- a/providers/flagd/pkg/provider.go
+++ b/providers/flagd/pkg/provider.go
@@ -473,7 +473,7 @@ func (p *Provider) handleEvents(ctx context.Context) error {
 			}
 			return err
 		case <-ctx.Done():
-			log.Info("Stop event handling with context done.")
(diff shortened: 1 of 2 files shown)


providers/flagd/pkg/provider.go
```diff
@@ -472,6 +472,10 @@ func (p *Provider) handleEvents(ctx context.Context) error {
 				WithoutCache()(p)
 			}
 			return err
+		case <-ctx.Done():
+			log.Info("Stop event handling with context done.")
```
skyerus, line 476:
```suggestion
			log.Info("stop event handling with context done")
```
Usual practice is lowercase without period

skyerus (approved):
nice :)

james-milligan (approved):

PR #92, opened 2023-01-11
chore: flagd provider cache invalidation test

### LINKED ISSUE #90: [flagd-provider] Cache invalidation unit testing
Implement missing cache invalidation unit testing.

skyerus:
## This PR

- Introduces unit testing of flagd-provider caching invalidation.
- Refactors provider struct to unexport Service (should not be exposed to application authors).
- Refactored provider tests to be in same package to allow use of unexported functions & manipulation of unexported fields (e.g. Service).
- Moved mock into it's own internal package (not exposed to application authors).

### Related Issues

Fixes #90


providers/flagd/pkg/provider.go
```diff
@@ -21,7 +21,7 @@ const defaultLRUCacheSize int = 1000
 
 type Provider struct {
 	ctx                              context.Context
-	Service                          service.IService
+	service                          service.IService
```
skyerus, line 24: This shouldn't be exported just for the sake of testing. Refactored tests into this same package to allow manipulation of this field.

Kavindu-Dodan:
This looks good and great job 👏 

Shall we also add a test to cover the error channel based cache invalidation ? [1]. You should be able to use the same approach and we can get it in with the same PR. 

[1] - https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/pkg/provider.go#L470

skyerus:
> This looks good and great job 👏
> 
> Shall we also add a test to cover the error channel based cache invalidation ? [1]. You should be able to use the same approach and we can get it in with the same PR.
> 
> [1] - https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/pkg/provider.go#L470

Thank you!
Great suggestion, done. @Kavindu-Dodan

james-milligan (approved):

Kavindu-Dodan (approved):

PR #112, opened 2023-02-10
feat: exposed WithTLS provider option. Allow tls to be used without cert path (default to host system's CAs)

### LINKED ISSUE #42: [flagd-provider] consistent options / TLS support without cert
We should make our options consistent, including the env vars. I think we want:

```
| Option name | Environment variable name | Type    | Default   |
| ----------- | ------------------------- | ------- | --------- |
| host        | FLAGD_HOST                | string  | localhost |
| port        | FLAGD_PORT                | number  | 8013      |
| tls         | FLAGD_TLS                 | boolean | false     |
| socketPath  | FLAGD_SOCKET_PATH         | string  | -         |
| certPath    | FLAGD_SERVER_CERT_PATH    | string  | -         |
```

We have all of these except `FLAGD_TLS`, which I think we want, in the case we want the provider to use TLS but also system certs instead of one made available at a specific path.

skyerus:
## This PR

- Exposes WithTLS provider options
- Allow TLS to be used without cert path (default to host system's CAs)

### Related Issues

Fixes #42

### DIFF
--- a/providers/flagd/pkg/provider.go
+++ b/providers/flagd/pkg/provider.go
@@ -17,7 +17,15 @@ import (
 	"strconv"
 )
 
-const defaultLRUCacheSize int = 1000
+const (
+	defaultLRUCacheSize                         int = 1000
+	flagdHostEnvironmentVariableName                = "FLAGD_HOST"
+	flagdPortEnvironmentVariableName                = "FLAGD_PORT"
+	flagdTLSEnvironmentVariableName                 = "FLAGD_TLS"
+	flagdSocketPathEnvironmentVariableName          = "FLAGD_SOCKET_PATH"
+	flagdServerCertPathEnvironmentVariableName      = "FLAGD_SERVER_CERT_PATH"
+	flagdCachingDisabledEnvironmentVariableName     = "FLAGD_CACHING_DISABLED"
+)
 
 type Provider struct {
 	ctx                              context.Context
@@ -34,6 +42,7 @@ type ProviderConfiguration struct {
 	Host            string
 	CertificatePath string
 	SocketPath      string
+	TLSEnabled      bool
 }
 
 type ProviderOption func(*Provider)
@@ -59,6 +68,7 @@ func NewProvider(opts ...ProviderOption) *Provider {
 			Port:            provider.providerConfiguration.Port,
 			CertificatePath: provider.providerConfiguration.CertificatePath,
 			SocketPath:      provider.providerConfiguration.SocketPath,
+			TLSEnabled:      provider.providerConfiguration.TLSEnabled,
 		},
 	}, provider.logger, nil)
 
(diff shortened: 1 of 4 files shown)


providers/flagd/pkg/provider.go
```diff
@@ -90,8 +100,7 @@ func WithSocketPath(socketPath string) ProviderOption {
 // FromEnv sets the provider configuration from environment variables: FLAGD_HOST, FLAGD_PORT, FLAGD_SERVICE_PROVIDER, FLAGD_SERVER_CERT_PATH & FLAGD_CACHING_DISABLED
 func FromEnv() ProviderOption {
 	return func(p *Provider) {
-
-		portS := os.Getenv("FLAGD_PORT")
+		portS := os.Getenv(flagdPortEnvironmentVariableName)
```
beeme1mr, line 103: Why is the S uppercase?

beeme1mr (approved):

providers/flagd/pkg/service/client.go
```diff
@@ -28,24 +27,27 @@ func (c *Client) Instance() schemaConnectV1.ServiceClient {
...
-				RootCAs: caCertPool,
+			tlsConfig = &tls.Config{}
+			if c.ServiceConfiguration.CertificatePath != "" {
+				caCert, err := os.ReadFile(c.ServiceConfiguration.CertificatePath)
+				if err != nil {
+					log.Fatal(err)
+				}
+				caCertPool := x509.NewCertPool()
+				if !caCertPool.AppendCertsFromPEM(caCert) {
+					log.Fatal("failed to AppendCertsFromPEM")
```
Kavindu-Dodan, line 48: Minor - may be a little bit of context for the error message ? IMO, the error log should highlight that provided certificates are malformed

providers/flagd/pkg/provider.go
```diff
@@ -17,7 +17,15 @@ import (
...
 
-const defaultLRUCacheSize int = 1000
+const (
+	defaultLRUCacheSize                         int = 1000
+	flagdHostEnvironmentVariableName                = "FLAGD_HOST"
+	flagdPortEnvironmentVariableName                = "FLAGD_PORT"
+	flagdTLSEnvironmentVariableName                 = "FLAGD_TLS"
+	flagdSocketPathEnvironmentVariableName          = "FLAGD_SOCKET_PATH"
+	flagdServerCertPathEnvironmentVariableName      = "FLAGD_SERVER_CERT_PATH"
+	flagdCachingDisabledEnvironmentVariableName     = "FLAGD_CACHING_DISABLED"
```
Kavindu-Dodan, line 27: Nice 👍

Kavindu-Dodan (approved):
Approving with minor change suggestion, which is opiniotted 👍

providers/flagd/pkg/provider.go
skyerus: I'm not sure tbh, I hadn't noticed the S. Happy to change it

providers/flagd/pkg/service/client.go
skyerus: Good shout :)

providers/flagd/pkg/provider.go
skyerus: S is for string in this instance, portString

beeme1mr: Okay, thanks. Must be a Go thing 😄

toddbaert (approved):

PR #176, opened 2023-04-19
feat: otel interceptor for flagd go-sdk

Kavindu-Dodan:
## This PR

Fulfill the sdk-side implementation of issue https://github.com/open-feature/flagd/issues/620

Introduce  option to enable disable otel interceptor for flagd connection from flagd go-sdk

SDK configuration - `openfeature.SetProvider(flagd.NewProvider(flagd.WithOtelInterceptor(true)))`

This option combined with open telemetry configurations at sdk usage component (ex:- application), allows distributed tracing for flagd 

## How to test

Write a simple code with otel traces and propagator, and use flagd with the option `flagd.WithOtelInterceptor(true)`

Related PR - https://github.com/open-feature/flagd/pull/624


providers/flagd/pkg/service/client_mock_test.go
```diff
@@ -1,64 +0,0 @@
-// Code generated by MockGen. DO NOT EDIT.
```
Kavindu-Dodan, line 1: removed as we do not use this

providers/flagd/Makefile
```diff
@@ -1,4 +1,3 @@
 .PHONY: mockgen
 mockgen:
-	mockgen -source=pkg/service/iservice.go -destination=internal/mock/service_mock_test.go -package=mock
-	mockgen -source=pkg/service/client.go -destination=pkg/service/client_mock_test.go -package=service_test
+	mockgen -source=pkg/service/iservice.go -destination=internal/mock/service_mock.go -package=mock
```
Kavindu-Dodan, line 3: This is a correction which we missed to update

thisthat (approved):

providers/flagd/README.md
```diff
@@ -22,6 +22,7 @@ flagd.WithoutCache()                    // disables caching of flag evaluations
 flagd.WithLRUCache(1000)                // enables LRU caching (see configuring caching section)
 flagd.WithBasicInMemoryCache()          // enables basic in memory cache (see configuring caching section)
 flagd.WithLogger(logger)                // sets a custom logger (see logging section)
+flagd.WithOtelInterceptor(bool)         // enable or disable OpenTelemetry interceptor for flagd communication
```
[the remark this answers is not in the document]
Kavindu-Dodan: Interceptor uses no-op implementations and yields nothing

The idea here was to allow sdk users to manually opt-in for telemetry interception. Because this is only needed if the sdk user (i.e- application developer) has a proper telemetry configuration for the application. Setup includes exporter configuration (ex:- console, otel exporter ...) and related services up and running (ex:- otel collector). 

So for simple usage as well as to exclude the performance impact of including no-op interceptors, it's best to make this an opt-in configuration.

providers/flagd/pkg/service/service.go
```diff
@@ -15,14 +15,6 @@ import (
 	"google.golang.org/protobuf/types/known/structpb"
 )
 
-type ServiceConfiguration struct {
```
Kavindu-Dodan, line 18: moved to `client.go` where this is properly used

providers/flagd/pkg/service/client.go
```diff
@@ -1,73 +1,104 @@
...
-			dialContext = func(_ context.Context, _, _ string) (net.Conn, error) {
-				return net.Dial("unix", c.ServiceConfiguration.SocketPath)
-			}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	if c.client != nil {
+		return c.client
+	}
+
```
toddbaert, line 48: I _think_ we can do the nil check first, and only lock if `c.client` is nil while we create it. That way, we only bother doing any locking if the `c.client` has not been created. Setting it is atomic, I think... so once it's been set and we unlock I guess we can never worry about locking again.

As is, this will lock with every access, which might not be needed. What do you think?

toddbaert (approved):
Approved, see my thought [here](https://github.com/open-feature/go-sdk-contrib/pull/176/files?diff=split&w=1#r1179581026).

providers/flagd/pkg/service/client.go
Kavindu-Dodan: After a second look at the code, I think the lock is an overkill. I will try to refactor the code to be bit nicer

Kavindu-Dodan: Refactored - no more locks and null checks. We have NewClient to generate the client [1]

[1] - https://github.com/open-feature/go-sdk-contrib/blob/9dc1bc714da8235dfb77a4e8934e0e6661d38e8d/providers/flagd/pkg/service/client.go

PR #181, opened 2023-04-25
feat: (GO Feature Flag provider) introducing cache in the provider

thomaspoignant:
## This PR
This PR introduces a TTL-based cache provider to the GO Feature Flag provider.

### Related Issues
Fixes https://github.com/thomaspoignant/go-feature-flag/issues/660


providers/go-feature-flag/pkg/provider.go
```diff
@@ -238,8 +300,36 @@ func evaluateLocally[T model.JsonType](provider *Provider, goffRequestBody model
 	}
 }
 
+func convertCache[T model.JsonType](value interface{}) model.GenericResolutionDetail[T] {
+	switch v := value.(type) {
+	case model.GenericResolutionDetail[T]:
+		return v
+	default:
+		// TODO: impossible to convert
+		panic("error")
```
james-milligan, line 309: this error could possibly result in the flushing of the cache, assuming the next call to `convertCache` wouldn't result in the same error on the next call, in which case it could result in the cache being disabled

providers/go-feature-flag/pkg/provider.go
```diff
@@ -59,17 +72,62 @@ func NewProvider(options ProviderOptions) (*Provider, error) {
...
+			"openfeature": "true",
+		},
+	}
+	if options.APIKey != "" {
+		webhook.Headers = map[string][]string{
+			"Authorization": {fmt.Sprintf("Bearer %s", options.APIKey)},
+		}
+	}
+
+	scheduler := exporter.NewScheduler(context.Background(),
```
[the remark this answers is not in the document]
thomaspoignant: I've added a new function called `NewProviderWithContext` to allow people to provide their context.

providers/go-feature-flag/pkg/provider.go
thomaspoignant: Good catch, I've done exactly what you said.

james-milligan (approved):

PR #209, opened 2023-05-15
feat: otel hook error status override option

### LINKED ISSUE #208: feat: option to override otel hook behavior
**Background**

With https://github.com/open-feature/go-sdk-contrib/issues/204, errors will set the span state to `Error`. However, a user might need to disable this behavior (ex:- as span is shared with logic other than flag evaluation). 

**Expected output**

Provide an option to disable the span error state.

Kavindu-Dodan:
## This PR

Fixes #208

This PR introduces hook build option `WithErrorStatusEnabled`. 

From README,

> **WithErrorStatusEnabled** : enable setting span status to `Error` in case of an error. Default behavior is disabled, 
  span status is unset for errors.

Code example

```
openfeature.AddHooks(
   otelhooks.NewHook(hooks.WithErrorStatusEnabled()))
```

This gives end-user the control over hook behavior

### DIFF
--- a/hooks/open-telemetry/README.md
+++ b/hooks/open-telemetry/README.md
@@ -13,8 +13,8 @@ A `context.Context` containing a `span` must be passed to the client evaluation 
 
 ### Options
 
-- WithErrorStatusDisabled: prevents setting span status to `Error` in case of an error. Default behavior is enabled, 
-  span status is set to `Error` if error occurs
+- WithErrorStatusEnabled: enable setting span status to `Error` in case of an error. Default behavior is disabled, 
+  span status is unset for errors.
 
 ### Example
 The following example demonstrates the use of the `OpenTelemetry hook` with the `OpenFeature go-sdk`. The traces are sent to a `zipkin` server running at `:9411` which will receive the following trace:
--- a/hooks/open-telemetry/README.md
+++ b/hooks/open-telemetry/README.md
@@ -11,6 +11,11 @@ For this hook to function correctly a global `TracerProvider` must be set, an ex
 The `open telemetry hook` taps into the after and error methods of the hook lifecycle to write `events` and `attributes` to an existing `span`.
 A `context.Context` containing a `span` must be passed to the client evaluation method, otherwise the hook will no-op.
 
+### Options
+
+- WithErrorStatusDisabled: prevents setting span status to `Error` in case of an error. Default behavior is enabled, 
+  span status is set to `Error` if error occurs
+
 ### Example
 The following example demonstrates the use of the `OpenTelemetry hook` with the `OpenFeature go-sdk`. The traces are sent to a `zipkin` server running at `:9411` which will receive the following trace:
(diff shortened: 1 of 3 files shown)


toddbaert:
@thisthat had some opinions about this (which mode should be the default). Personally I'm fine with marking the span as an error by default, but I'd like to come to an agreement, because we should be consistent with all implementations. I'm also not an OTel expert by any stretch.

cc @bacherfl @beeme1mr @thisthat

thisthat:
There are arguments for both directions. I suggest recording an OpenFeature error because a FF evaluation constitutes an exception that is internally handled and does not impact the "client request". Marking the Span (and consequentially the whole trace) as an error could potentially trigger alerts on the Observability platform used to ingest traces.

From the [OTel spec](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/api.md#set-status) they say the following

> Analysis tools SHOULD respond to an Ok status by suppressing any errors they would otherwise generate. For example, to suppress noisy errors such as 404s.

I would consider a FF evaluation under this category. 
Nevertheless, OTel also has some recommendations handy-dandy APIs to record exception, which fit our use case imho: https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/exceptions.md#recording-an-exception

Since the internal OF error is never pushed to the user, I would suggest recording the event with `exception.escaped = false`. The SDK should already handle all the remaining properties (see https://github.com/open-telemetry/opentelemetry-go/blob/main/sdk/trace/span.go#L427)

Kavindu-Dodan:
@thisthat  thank you for the review and references.

The current hook implementation performs the error recording [1]. And the newest change here was to set the status to `Error`.

This PR gives the end user the **option** to **disable** setting the Span status to `Error`. And this could help them with error suppression if that's the desired behavior. 

So can we agree on the following, 

- Hook will record & by default set the status to `Error` for OpenFeature errors
- Hook will provide configurability to disable the status setting behavior (i.e - disable setting Span status to `Error`)


[1] - https://github.com/open-feature/go-sdk-contrib/blob/main/hooks/open-telemetry/pkg/otel.go#L41-L44

thisthat:
Thanks @Kavindu-Dodan for the answer. I have the feeling that we should invert the default behavior, i.e., by default not marking the request as erroneous. I think in the majority of cases, I don't want to mark a request failed because there was a problem in the FF evaluation, for the end user everything worked fine.

Kavindu-Dodan:
@thisthat I have updated the PR to support the agreement. 

Plus, updated PR description & underlying issue for visibility. In summary, 

- Hooks will **NOT** set Span status to`Error` by default
- The end user will get an **option** (option name - `WithErrorStatusEnabled`) to configure the hook to set `Error` status for errors 

cc - @toddbaert @beeme1mr @bacherfl

thisthat (approved):

beeme1mr (approved):

PR #217, opened 2023-05-19
feat: metric hooks

Kavindu-Dodan:
## This PR

Attempt to introduce Metric Hooks for go-sdk. This is **pre-ofep** POC, hence marked as draft state 

### Why metrics ? 

Current OpenTelemetry hook focus on Spans. But, metrics are also key telemetry details. 

### How ? 

OpenFeature hook stages (before, after, error & finally) [1] allow clear points to interact and collect metrics. In this POC, we collect 4 metrics using those stages, 

- **evaluationActive**: Active evaluations under SDK. Collected through an `updowncounter` [2], utilizing `before` & `finally` hook stages
- **evaluationRequests**: Number of flag evaluation requests SDK received. Collected at `before` stage and using a `Counter` [3]
- **evaluationSuccess**: Successful evaluations. Collected at `after` stage and using a `Counter` [3]
- **evaluationError**: Unsuccessful evaluations. Collected at `finally` stage and using a `Counter` [3]

All above metrics are collected with dimensions agreed with OpenTelemeletry semantic conventions for feature flags [4]

---- 

Next steps 

- [ ] OFEP for community feedback
- [ ] Improve this implementation with tests & community feedback
- [ ] Implementation for other language SDKs

[1] - https://openfeature.dev/specification/sections/hooks
[2] - https://opentelemetry.io/docs/specs/otel/metrics/api/#updowncounter
[3] - https://opentelemetry.io/docs/specs/otel/metrics/api/#counter 
[4] - https://opentelemetry.io/docs/specs/otel/logs/semantic_conventions/feature-flags/


beeme1mr:
We should also consider restricting the dimensions of a failed evaluation. For example, you likely won't want to use flag key if the error is "flag not found".

Kavindu-Dodan:
> What's the use case for evaluation active?

The idea here is to gather active evaluations within the SDK. I think this should help to see bottlenecks if any under high load scenarios. However, as described at ofep [1] for cache based or fast evaluations, we should observe a `0` value here.

> We should also consider restricting the dimensions of a failed evaluation. For example, you likely won't want to use flag key if the error is "flag not found".

I think this is true from a provider perspective. However, the SDK is used by an application, so there is only a finite number of flag keys. 

Anyway we can discuss this further through OFEP [2]

[1] - https://github.com/open-feature/ofep/pull/62/files#diff-1ee0937abd11c52172d0be2b854ab5162b15262da53e7ddbcd4090484879bc67R35-R38
[2] -  https://github.com/open-feature/ofep/pull/62

hooks/open-telemetry/pkg/metrics.go
```diff
@@ -0,0 +1,100 @@
+package otel
```
toddbaert, line 1: Hmmm... what you did here was _not_ introduce another package, but add the hook to the existing package... I never considered that solution honestly, in my mind, we'd add a totally new package that would be a different dependency.

I think there's merit to this approach though, in this repo as well as others. We could put the metrics and traces hooks in the same artifact (hooks.otel) and just provide 2 different hooks in that jar/module, etc.

Perhaps the only downside would be that the metrics AND the traces OTel deps would be bundled - but I don't even know if that's much of an issue since they would be tree shaken or not compiled in many cases.

toddbaert: @beeme1mr see above.

beeme1mr: Seems reasonable to me. At least in Go.

beeme1mr (commented):
Please update the hook readme to include usage information.

hooks/open-telemetry/pkg/metrics.go
toddbaert: @Kavindu-Dodan I think we may want to rename `otel.go` and `NewHook` to `traces.go` and `NewTraceHook` respectively. Yes it will be a breaking change, but we are <0 and it's a very easy change to absorb.

I dont mind whether it's done in this PR or another, but IMO we should do this before we release this new stuff.

hooks/open-telemetry/README.md
```diff
@@ -1,38 +1,92 @@
...
+- `feature_flag.evaluation_active_count`
+
+### Options
+
+#### WithFlagMetadataDimensions 
+
+This constructor option allows to configure dimension descriptions to be extracted from `openfeature.FlagMetadata`. 
+If present, these dimension will be added to the `feature_flag.evaluation_success_total` metric.
+
+Note that, this configuration must be carefully coordinated with `Provider` as missing dimensions results in hook
```
beeme1mr, line 26: Does that mean the hook will through if `scope` is missing from the hook metadata?

hooks/open-telemetry/README.md
```diff
@@ -1,38 +1,92 @@
...
+}
+
+// Register OpenFeature API level hooks
+openfeature.AddHooks(metricsHook)
+```
+
+## Span hook
+
+For this hook to function correctly a global `TracerProvider` must be set, an example of how to do this can be found
+below.
```
beeme1mr, line 59: Is your IDE automatically adding line breaks based on the sentence length?

hooks/open-telemetry/README.md
```diff
@@ -1,38 +1,92 @@
...
 
 - open-feature/go-sdk >= v1.3.0
 
 ## Usage
 
-For this hook to function correctly a global `TracerProvider` must be set, an example of how to do this can be found below.
+## Metric hook
 
-The `open telemetry hook` taps into the after and error methods of the hook lifecycle to write `events` and `attributes` to an existing `span`.
+This hook performs metric collection by tapping into various hook stages. Given below are the metrics are extracted
```
beeme1mr, line 11: Please do a line break at a new sentence instead of after a character count (where possible).

hooks/open-telemetry/README.md
Kavindu-Dodan: Yeah, this is from formatting. However, this does not impact the final rendering, see here - https://github.com/open-feature/go-sdk-contrib/blob/30035447e7d114cf2052fed669897cc57f67154d/hooks/open-telemetry/README.md 

IMO, it's good to have a character limit per line and add a line break. However, in .MD rendering this does not necessarily convert to a new line

hooks/open-telemetry/README.md
Kavindu-Dodan: By default (no configurations) nothing will happen. See this test case as an example [1]. However, if someone configures the hook with  `WithFlagMetadataDimensions`, similar example below,

```go
NewMetricsHook(reader,
    WithFlagMetadataDimensions(
        DimensionDescription{
            Key:  "scope",
            Type: String,
        }))
```

then the logic will try to extract the flag metadata entry, which returns an error. And this error comes from go-sdk itself , see GetString for example [2]

[1] - https://github.com/open-feature/go-sdk-contrib/pull/217/files#diff-4947cec855aa6f6fb207103527911ad05567255efcc5f569bbcc0d3ffe80a17fR57-R109
[2] - https://github.com/open-feature/go-sdk/blob/main/pkg/openfeature/client.go#L176

hooks/open-telemetry/README.md
beeme1mr: Adding a new line is totally fine but it would make providing feedback easier if it was split per sentence.

hooks/open-telemetry/pkg/metrics.go
Kavindu-Dodan: Yeah agree. This can be a follow-up PR, but should prior to the release as you said :)

Kavindu-Dodan: Created a follow-up issue https://github.com/open-feature/go-sdk-contrib/issues/224

beeme1mr (commented):
I think scope should be captured automatically if it's set as flag metadata. That would also mean the hook shouldn't error if it isn't present. The option could be used to opt out of capturing if that's the desired behavior.

This seems like the desired behavior because the flag metadata will only be sent if the provider deems it valuable. Also, having the hook error if the property isn't present may prove challenging in multi-provider environments.

Kavindu-Dodan:
> I think scope should be captured automatically if it's set as flag metadata. That would also mean the hook shouldn't error if it isn't present. The option could be used to opt out of capturing if that's the desired behavior.
> 
> This seems like the desired behavior because the flag metadata will only be sent if the provider deems it valuable. Also, having the hook error if the property isn't present may prove challenging in multi-provider environments.

The problem is with the dimension naming. If we need the default opt-in behavior, we need to define the dimension name through the spec/glossary. Current OFEP highlights this dimension [1] but does not enforce a naming. And the implementation allows extracting any dimension bound to flag evaluation metadata. I believe flexibility in dimension extraction is better than strictly focusing on a single dimension (a generic implementation). 

Regarding error handling - the main issue is the lack of a logger (log and continue VS throw error). Making the hook accept a generic logger is not straightforward at least in Go

[1] - https://github.com/open-feature/ofep/pull/62/files#diff-1ee0937abd11c52172d0be2b854ab5162b15262da53e7ddbcd4090484879bc67R77-R85

hooks/open-telemetry/README.md
Kavindu-Dodan: Updated the implementation to ignore missing metadata and avoid error returns. I think this is the least complex implementation.

thisthat (approved):
LGTM beside a nitpick 👍

hooks/open-telemetry/pkg/metrics_test.go
```diff
@@ -0,0 +1,405 @@
...
+
+	if len(instrument.DataPoints) < 1 {
+		t.Error("expected data points, but found none")
+	}
+
+	attributes := instrument.DataPoints[0].Attributes
+
+	value, ok := attributes.Value(attribute.Key(scopeDescription.Key))
+	if !ok || value.AsString() != scopeValue {
+		t.Errorf("attribute %s is incorrectoly configured", scoreDescription.Key)
```
RealAnna, line 375: I know it is a nit but there is a repeated typo here :) 
```suggestion
		t.Errorf("attribute %s is incorrectly configured", scoreDescription.Key)
```

Kavindu-Dodan: Really good observation 🙇

toddbaert (approved):

PR #237, opened 2023-06-08
chore: rename constructor method

### LINKED ISSUE #224: [chore] Rename otel hooks to NewTraceHook
The current hook constructor is too generic `NewHook` [1], that it doesn't convey proper idea behind the hook.

This should be renamed to something meaningful like `NewTraceHook`

[1] - https://github.com/open-feature/go-sdk-contrib/blob/main/hooks/open-telemetry/pkg/otel.go#L25

Kavindu-Dodan:
## This PR

Fixes #224

Introduced `NewTracesHook` and deprecated `NewHook` to construct the OTel hook. This is not a breaking change as old constructor method is preserved.

### DIFF
--- a/hooks/open-telemetry/pkg/traces.go
+++ b/hooks/open-telemetry/pkg/traces.go
@@ -16,14 +16,21 @@ const (
 	EventPropertyVariant      = "feature_flag.variant"
 )
 
-type hook struct {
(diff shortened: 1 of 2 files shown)


james-milligan (approved):

toddbaert:
This was a good way to do this. No breaking change is nice.

PR #250, opened 2023-06-27
feat: Return CACHED reason when using cache

thomaspoignant:
## This PR
- Return `CACHED` as a reason when using the cache in the provider.


james-milligan (approved):
LGTM

PR #278, opened 2023-07-21
feat: flag evaluation metadata

Kavindu-Dodan:
## This PR

Related to flagd change at https://github.com/open-feature/flagd/pull/761

Handles evaluation metadata from flagd and fills required OF SDK field


thisthat (commented):
the changes look good, I would suggest to for at least one test entry some metadata to protect against regression while caching it. Similarly, we can add some unit test for the evaluation methods in `provider_test.go`

Kavindu-Dodan:
> the changes look good, I would suggest to for at least one test entry some metadata to protect against regression while caching it. Similarly, we can add some unit test for the evaluation methods in `provider_test.go`

Thank you for the review. Updated boolean eval tests to have metadata - https://github.com/open-feature/go-sdk-contrib/pull/278/commits/f6274c2797032e7051237e6d57d4391d53de6454

toddbaert (approved):

PR #282, opened 2023-07-25
fix: go-feature-flag sdk - cache key

liran2000:
Problem:
Flag evaluation for same flag using different evaluation context properties can lead to cache key collision.
Solution:
Build cache key considering flag and user evaluation context properties.

### Related Issues
[java-sdk PR 369](https://github.com/open-feature/java-sdk-contrib/pull/369)

### How to test
Run unit tests.
Can test via go-feature-flag instance, while observing logs.

### DIFF
--- a/providers/go-feature-flag/pkg/provider.go
+++ b/providers/go-feature-flag/pkg/provider.go
@@ -317,7 +317,7 @@ func convertCache[T model.JsonType](value interface{}) (model.GenericResolutionD
 
 // evaluateWithRelayProxy is calling GO Feature Flag relay proxy to evaluate the file.
 func evaluateWithRelayProxy[T model.JsonType](provider *Provider, ctx context.Context, goffRequestBody model.EvalFlagRequest, flagName string, defaultValue T) model.GenericResolutionDetail[T] {
-	cacheKey := fmt.Sprintf("%s-%+v", flagName, goffRequestBody.User)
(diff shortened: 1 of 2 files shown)


providers/go-feature-flag/pkg/provider.go
```diff
@@ -317,7 +317,7 @@ func convertCache[T model.JsonType](value interface{}) (model.GenericResolutionD
 
 // evaluateWithRelayProxy is calling GO Feature Flag relay proxy to evaluate the file.
 func evaluateWithRelayProxy[T model.JsonType](provider *Provider, ctx context.Context, goffRequestBody model.EvalFlagRequest, flagName string, defaultValue T) model.GenericResolutionDetail[T] {
-	cacheKey := fmt.Sprintf("%s-%s", flagName, goffRequestBody.User.Key)
+	cacheKey := fmt.Sprintf("%s-%+v", flagName, goffRequestBody.User)
```
thomaspoignant, line 320:
```suggestion
	cacheKey := fmt.Sprintf("%s-%+v", flagName, goffRequestBody.EvaluationContext)
```

This is my mistake here, but we should better use `EvaluationContext` than user, because `User` should be deprecated.

thomaspoignant (approved):

thomaspoignant:
CI is failing on lint because of this issue https://github.com/open-feature/go-sdk-contrib/issues/287

PR #311, opened 2023-08-08
feat!: attribute setter callbacks for otel hooks and remove deprecated constructors 

### LINKED ISSUE #294: [otel hooks] add functional means of adding custom attributes to metrics and tracing hooks
Currently it's possible to map flag metadata values to custom OTel attributes (see [here](https://github.com/open-feature/go-sdk-contrib/tree/main/hooks/open-telemetry#withflagmetadatadimensions)). However this approach has limitations, as it can only map values 1:1.

We should implement something similar to the [JS implementation](https://github.com/open-feature/js-sdk-contrib/tree/main/libs/hooks/open-telemetry#custom-metric-attributes), which allows consumers to provide a function that takes the flag metadata and returns OTel attributes.

We should add the same functionality to the Tracing hook.

Kavindu-Dodan:
## This PR

FIxes #294

Adds new OTel hook constructor options for `TraceHook` & `MetricsHooks`

New options `WithAttributeSetterForTraces` & `WithAttributeSetterForMetrics` allow the registration of custom callbacks to extract OTel attributes (dimensions) from flag evaluation metadata 

Callback has the following signature,

```Go
func(openfeature.FlagMetadata) []attribute.KeyValue
```

Along with this change, I have removed deprecated constructors to fascilitate easy future enahancements

### DIFF
--- a/hooks/open-telemetry/pkg/metrics.go
+++ b/hooks/open-telemetry/pkg/metrics.go
@@ -25,7 +25,7 @@ type MetricsHook struct {
 	errorCounter   api.Int64Counter
 
 	flagEvalMetadataDimensions []DimensionDescription
-	attributeSetterCallback    *func(openfeature.FlagMetadata) []attribute.KeyValue
+	attributeMapperCallback    *func(openfeature.FlagMetadata) []attribute.KeyValue
 }
 
 var _ openfeature.Hook = &MetricsHook{}
@@ -98,8 +98,8 @@ func (h *MetricsHook) After(ctx context.Context, hCtx openfeature.HookContext,
 	fromMetadata := descriptionsToAttributes(details.FlagMetadata, h.flagEvalMetadataDimensions)
 	attribs = append(attribs, fromMetadata...)
 
-	if h.attributeSetterCallback != nil {
-		attribs = append(attribs, (*h.attributeSetterCallback)(details.FlagMetadata)...)
+	if h.attributeMapperCallback != nil {
+		attribs = append(attribs, (*h.attributeMapperCallback)(details.FlagMetadata)...)
 	}
 
 	h.successCounter.Add(ctx, 1, api.WithAttributes(attribs...))
@@ -149,11 +149,11 @@ func WithFlagMetadataDimensions(descriptions ...DimensionDescription) MetricOpti
 	}
 }
 
-// WithAttributeSetterForMetrics allows to set a extractionCallback which accept openfeature.FlagMetadata and returns
+// WithMetricsAttributeSetter allows to set a extractionCallback which accept openfeature.FlagMetadata and returns
 // []attribute.KeyValue derived from those metadata.
-func WithAttributeSetterForMetrics(callback *func(openfeature.FlagMetadata) []attribute.KeyValue) MetricOptions {
+func WithMetricsAttributeSetter(callback *func(openfeature.FlagMetadata) []attribute.KeyValue) MetricOptions {
 	return func(metricsHook *MetricsHook) {
-		metricsHook.attributeSetterCallback = callback
+		metricsHook.attributeMapperCallback = callback
 	}
 }
 
--- a/hooks/open-telemetry/pkg/metrics.go
+++ b/hooks/open-telemetry/pkg/metrics.go
@@ -25,18 +25,10 @@ type MetricsHook struct {
 	errorCounter   api.Int64Counter
 
 	flagEvalMetadataDimensions []DimensionDescription
+	attributeSetterCallback    *func(openfeature.FlagMetadata) []attribute.KeyValue
 }
 
-type MetricOptions func(*MetricsHook)
-
-// NewMetricsHook builds a metric hook backed by provided metric.Reader. Reader must be provided by developer and
-// its configurations govern metric exports
-// Deprecated: NewMetricsHook is deprecated. Please use NewMetricsHookFromProvider
-func NewMetricsHook(reader metric.Reader, opts ...MetricOptions) (*MetricsHook, error) {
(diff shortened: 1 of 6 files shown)


hooks/open-telemetry/pkg/metrics.go
```diff
@@ -149,6 +149,14 @@ func WithFlagMetadataDimensions(descriptions ...DimensionDescription) MetricOpti
 	}
 }
 
+// WithAttributeSetterForMetrics allows to set a extractionCallback which accept openfeature.FlagMetadata and returns
+// []attribute.KeyValue derived from those metadata.
```
thisthat, line 153:
```suggestion
// []attribute.KeyValue derived from that metadata.
```

hooks/open-telemetry/pkg/traces.go
```diff
@@ -81,4 +83,11 @@ func WithErrorStatusEnabled() Options {
 	}
 }
 
-var _ openfeature.Hook = &traceHook{}
+// WithAttributeSetterForTraces allows to set a extractionCallback which accept openfeature.FlagMetadata and returns
+// []attribute.KeyValue derived from those metadata.
```
thisthat, line 87:
```suggestion
// []attribute.KeyValue derived from that metadata.
```

thisthat (approved):
left two nit picts

hooks/open-telemetry/pkg/metrics.go
```diff
@@ -25,18 +25,10 @@ type MetricsHook struct {
 	errorCounter   api.Int64Counter
 
 	flagEvalMetadataDimensions []DimensionDescription
+	attributeSetterCallback    *func(openfeature.FlagMetadata) []attribute.KeyValue
```
toddbaert, line 28: nit: `attributeMapperCallback` or `attributeMapper` might be better, but that's totally up to you.

hooks/open-telemetry/pkg/metrics.go
```diff
@@ -149,6 +149,14 @@ func WithFlagMetadataDimensions(descriptions ...DimensionDescription) MetricOpti
 	}
 }
 
+// WithAttributeSetterForMetrics allows to set a extractionCallback which accept openfeature.FlagMetadata and returns
+// []attribute.KeyValue derived from that metadata.
+func WithAttributeSetterForMetrics(callback *func(openfeature.FlagMetadata) []attribute.KeyValue) MetricOptions {
```
toddbaert, line 154: Maybe `WithTracesAttributeSetter` or `WithTracesAttributeMapper`.

hooks/open-telemetry/pkg/traces.go
```diff
@@ -81,4 +83,11 @@ func WithErrorStatusEnabled() Options {
 	}
 }
 
-var _ openfeature.Hook = &traceHook{}
+// WithAttributeSetterForTraces allows to set a extractionCallback which accept openfeature.FlagMetadata and returns
+// []attribute.KeyValue derived from that metadata.
+// If present, returned attributes will be added to successful evaluation traces
+func WithAttributeSetterForTraces(callback *func(openfeature.FlagMetadata) []attribute.KeyValue) Options {
```
toddbaert, line 89: see https://github.com/open-feature/go-sdk-contrib/pull/311/files#r1291638254

hooks/open-telemetry/pkg/metrics.go
toddbaert, line 28: Think we need to support returning errors from the function and logging? I'm not sure, totally up to you if you think that would be valuable enough to implement.

toddbaert (approved):
Only optional, mostly naming nits.

hooks/open-telemetry/pkg/metrics.go
Kavindu-Dodan: Changed to `attributeMapperCallback`

hooks/open-telemetry/pkg/metrics.go
Kavindu-Dodan: Renamed to `WithMetricsAttributeSetter`

hooks/open-telemetry/pkg/traces.go
Kavindu-Dodan: Renamed to `WithTracesAttributeSetter`

hooks/open-telemetry/pkg/metrics.go
Kavindu-Dodan: I don't think this is needed. Having an error in the callback creates a contract that we need to take care of from SDK end 🤔  IMO this is not needed and error handling must be done by the callback writer.

hooks/open-telemetry/pkg/metrics.go
```diff
@@ -25,18 +25,10 @@ type MetricsHook struct {
 	errorCounter   api.Int64Counter
 
 	flagEvalMetadataDimensions []DimensionDescription
+	attributeMapperCallback    *func(openfeature.FlagMetadata) []attribute.KeyValue
```
craigpastro, line 28: Hello 👋 I hope you don't mind me commenting on your PR, but I don't think you need to use a pointer here and it may be cleaner without it. The zero value of a function is nil.

Kavindu-Dodan: Thank you @craigpastro  for the review 

I updated the changes to accept nonpointer functions - https://github.com/open-feature/go-sdk-contrib/pull/311/commits/4c8b3d1e51b4cc3da09acc3895ec7310a8560b72

My initial reasoning for this was to have a comparability of the callback. But I think this is never going to be needed as we only have one callback registration. 

Also, thank you for mentioning documentation change need. I updated the readme accordingly

PR #317, opened 2023-08-25
feat: Eventing support for flagd provider

Kavindu-Dodan:
## This PR

Adds OpenFeature Events extension [^1] support to flagd provider

**tldr**

- Use`IService.go` as the main interface between provider and grpc handling (flagd connectivity)
- Cache is now extracted to a dedicated service (potential improvements exist here)
- Provider configurations are extracted to a dedicated component named `configuration.go`
- Eventing support 

Review commit by commit OR with a checkout of the original branch to have more focus 

----

Main refactoring on this PR focuses on isolating grpc based flagd connectivity to the service layer. This was done by refactoring `provider.go` and updating `IService.go`. With the change, we no longer have grpc dependencies at provider level. This is favorable for future efforts such as in-memory provider capability proposed through https://github.com/open-feature/flagd/pull/848 

Caching is now isolated to a dedicated service. This isolation helps to pass the service to appropriate caching locations and maintain a single cache within the provider. 

Provider configurations were also extracted to a new component to reduce logic at the provider level. 

Given below is an overview of component relationships in the current state, 

image

[^1]: https://openfeature.dev/specification/sections/events


Makefile
```diff
@@ -12,7 +12,7 @@ test:
 	go list -f '{{.Dir}}/...' -m | xargs -I{} go test -v {}
 
 lint: 
-	go install -v github.com/golangci/golangci-lint/cmd/golangci-lint@latest
+	go install -v github.com/golangci/golangci-lint/cmd/golangci-lint@v1.54.1
```
Kavindu-Dodan, line 15: the latest release is not compatible with go version 1.19+ variants

providers/flagd/README.md
```diff
@@ -44,16 +44,16 @@ func main() {
...
-| maxCacheSize          | FLAGD_MAX_CACHE_SIZE           | int       |              | 1000      |
-| maxEventStreamRetries | FLAGD_MAX_EVENT_STREAM_RETRIES | int       |              | 5         |
+| Option name           | Environment variable name      | Type    | Options            | Default   |
+|-----------------------|--------------------------------|---------|--------------------|-----------|
+| host                  | FLAGD_HOST                     | string  |                    | localhost |
+| port                  | FLAGD_PORT                     | number  |                    | 8013      |
+| tls                   | FLAGD_TLS                      | boolean |                    | false     |
+| socketPath            | FLAGD_SOCKET_PATH              | string  |                    |           |
+| certPath              | FLAGD_SERVER_CERT_PATH         | string  |                    |           |
+| cache                 | FLAGD_CACHE                    | string  | lru, mem, disabled | lru       |
```
toddbaert, line 54: Can you explain why you added this new cache type?

I guess it's not a problem, but I wonder why anyone would use it over LRU.

providers/flagd/pkg/provider.go
```diff
@@ -1,213 +1,164 @@
...
-// WithPort specifies the port of the flagd server. Defaults to 8013
+// WithPort specifies the expectPort of the flagd server. Defaults to 8013
 func WithPort(port uint16) ProviderOption {
 	return func(p *Provider) {
 		p.providerConfiguration.Port = port
 	}
 }
 
-// WithHost specifies the host name of the flagd server. Defaults to localhost
+// WithHost specifies the expectHost name of the flagd server. Defaults to localhost
```
toddbaert, line 161:
```suggestion
// WithHost specifies the host name of the flagd server. Defaults to localhost
```
guessing this is a typo

providers/flagd/pkg/provider.go
```diff
@@ -1,213 +1,164 @@
...
+// If certificate loading fails insecure credentials will be used instead
 func WithCertificatePath(path string) ProviderOption {
 	return func(p *Provider) {
 		p.providerConfiguration.CertificatePath = path
 		p.providerConfiguration.TLSEnabled = true
 	}
 }
 
-// WithPort specifies the port of the flagd server. Defaults to 8013
+// WithPort specifies the expectPort of the flagd server. Defaults to 8013
```
toddbaert, line 154:
```suggestion
// WithPort specifies the port of the flagd server. Defaults to 8013
```
guessing this is a typo

providers/flagd/pkg/provider.go
```diff
@@ -1,213 +1,164 @@
...
-	SocketPath      string
-	TLSEnabled      bool
-}
+	logger                logr.Logger
+	providerConfiguration *providerConfiguration
+	service               IService
+	status                of.State
 
-type ProviderOption func(*Provider)
+	eventStream chan of.Event
```
toddbaert, line 18: channels "feel" great for this.

providers/flagd/pkg/service/service.go
```diff
@@ -50,131 +76,292 @@ type resolutionResponseConstraints interface {
...
 
-	return res.Msg, nil
+	go func() {
+		s.startEventStream(ctx)
+	}()
+
+	s.cancelHook = cancelFunc
+
+	return nil
+}
```
toddbaert, line 95: Unless I'm mistaken, we have a similar issue here as we did in Java, where we don't know for sure that the event stream is successfully connected when `Init` is complete.

Can we block here until a ready event?

providers/flagd/pkg/service/service.go
```diff
@@ -50,131 +76,292 @@ type resolutionResponseConstraints interface {
...
+
+	if s.cache.IsEnabled() {
+		fromCache, ok := s.cache.GetCache().Get(key)
+		if ok {
+			fromCacheResDetail, ok := fromCache.(openfeature.BoolResolutionDetail)
+			if ok {
+				fromCacheResDetail.Reason = ReasonCached
+				return fromCacheResDetail
+			}
+		}
```
toddbaert, line 113: It would be great if we could use generics or some other mechanism to reduce this duplication.

providers/flagd/pkg/service/service.go
```diff
@@ -1,40 +1,66 @@
 package service
 
 import (
+	schemaConnectV1 "buf.build/gen/go/open-feature/flagd/bufbuild/connect-go/schema/v1/schemav1connect"
```
toddbaert, line 4: This is a nit, but now that you've isolated most of the gRPC code here, you may consider calling this file grpc_service... not necessary though.

If you do, I would recommend you do it with `git mv ...` so that it's easier to see in review.

providers/flagd/README.md
Kavindu-Dodan: We always supported in-memory cache but lacked documentation under environment configuration section - https://github.com/open-feature/go-sdk-contrib/blob/6545b4c16db57114456c58a3eb2966ec461ae9f9/providers/flagd/README.md#using-flagdfromenv

providers/flagd/pkg/provider.go
Kavindu-Dodan: 🤦 this is from a test variable renaming. Thank you for noticing

providers/flagd/pkg/provider.go
Kavindu-Dodan: yup, this is a rename ALL mistake 

original usage - https://github.com/open-feature/go-sdk-contrib/blob/b777a3d9efe858be0eacf26056d31cf9debec444/providers/flagd/pkg/provider_test.go#L13-L14

providers/flagd/pkg/provider.go
Kavindu-Dodan: Yeah, emit the event and let SDK handle it. IMO this is much better than shared objects + locks

providers/flagd/pkg/service/service.go
Kavindu-Dodan: Good suggestion. But I would like to defer this renaming to a follow-up PR. This one already has quite a lot of changes (IMO too many for a single pr 😞 ) so changing the packaging could confuse you even more.

providers/flagd/pkg/service/service.go
Kavindu-Dodan: 🤦 this is something I always forget about the spec.

I fixed this with the latest commit - https://github.com/open-feature/go-sdk-contrib/pull/317/commits/607ba6e84fcf4b2c278d65a13706ba3eba6518da 

however, this also shows there's some corner cases when it comes to handling events 🤔 something we could improve

providers/flagd/pkg/service/service.go
Kavindu-Dodan: +1 on this. I did not attempt this to reduce review efforts. Change in the PR fully maps to current main's logic. But once merged, we might be able to find some commons and extract them to generic handlers

providers/flagd/pkg/configuration.go
```diff
@@ -0,0 +1,116 @@
+package flagd
+
+import (
+	"fmt"
+	"github.com/go-logr/logr"
+	"github.com/open-feature/go-sdk-contrib/providers/flagd/pkg/cache"
```
thisthat, line 6: I think these are not formatted correctly

providers/flagd/pkg/configuration.go
```diff
@@ -0,0 +1,116 @@
...
+		EventStreamConnectionMaxAttempts: defaultMaxEventStreamRetries,
+		Host:                             defaultHost,
+		MaxCacheSize:                     defaultMaxCacheSize,
+		Port:                             defaultPort,
+		TLSEnabled:                       defaultTLS,
+	}
+}
+
+// updateFromEnvVar is a utility to update configurations based on current environment variables
+func (cfg *providerConfiguration) updateFromEnvVar(logger logr.Logger) {
```
thisthat, line 56: I see that the logger is not part of `providerConfiguration` but it feels weird to have it as an argument of a function since it could vary between calls to this function.

providers/flagd/pkg/configuration.go
```diff
@@ -0,0 +1,116 @@
...
+				))
+		} else {
+			cfg.Port = uint16(port)
+		}
+	}
+
+	host := os.Getenv(flagdHostEnvironmentVariableName)
+	if host != "" {
+		cfg.Host = host
+	}
```
thisthat, line 74: nit. This way we scope `host` to the bare minimum
```suggestion
	if host := os.Getenv(flagdHostEnvironmentVariableName); host != "" {
		cfg.Host = host
	}
```

providers/flagd/pkg/configuration.go
```diff
@@ -0,0 +1,116 @@
...
+
+	socketPath := os.Getenv(flagdSocketPathEnvironmentVariableName)
+	if socketPath != "" {
+		cfg.SocketPath = socketPath
+	}
+
+	certificatePath := os.Getenv(flagdServerCertPathEnvironmentVariableName)
+	if certificatePath != "" || os.Getenv(flagdTLSEnvironmentVariableName) == "true" {
```
thisthat, line 82: maybe worth trimming and lowercase the value of `flagdTLSEnvironmentVariableName`

providers/flagd/pkg/service/service.go
```diff
@@ -193,66 +380,173 @@ func handleError(err error) openfeature.ResolutionError {
...
-			delay = s.baseRetryDelay // reset delay if the connection was successful before failing
-		}
-		if err != nil && i <= maxAttempts {
+func (s *Service) EventChannel() <-chan of.Event {
+	return s.events
+}
+
+// startEventStream - starts listening to flagd event stream with retries.
+// If retrying is exhausted, an event with openfeature.ProviderError is emitted.
+func (s *Service) startEventStream(ctx context.Context) {
```
thisthat, line 389: maybe worth adding a line saying that this function is expected to be called in a go routine

providers/flagd/pkg/service/service.go
```diff
@@ -193,66 +380,173 @@ func handleError(err error) openfeature.ResolutionError {
...
+	var delay = s.baseRetryDelay
+	var attempts = 1
+
+	// wraps connection with retry attempts
+	for attempts <= s.maxRetries {
+		s.logger.V(logger.Debug).Info("connecting to event stream")
+		err := s.streamClient(ctx)
```
thisthat, line 396: I think here we missed resetting the connection attempts if we successfully connect.

providers/flagd/pkg/service/service.go
toddbaert: Yep, that looks good to me.

providers/flagd/pkg/configuration.go
Kavindu-Dodan: Updated with the commit - https://github.com/open-feature/go-sdk-contrib/pull/317/commits/158f552c75623a69b753f67afb835e14b4ddbeaf

I made logger part of the config instance creation so that we don't have to call it with a logger

providers/flagd/pkg/configuration.go
Kavindu-Dodan: Updated all if conditions to use the same format

providers/flagd/pkg/service/service.go
Kavindu-Dodan: This handled at the else block [^1]

And error handling the stream means there's something wrong, which we should retry. This could be a connection loss, packet errors or something else.

However, if there's no error, which can happen with a provider shutdown, stream end, we reset attempts and starts fresh connection establishment attempts 

[^1]: https://github.com/open-feature/go-sdk-contrib/blob/470ee860a64ccade08b6e764962e5b9d50c0a210/providers/flagd/pkg/service/service.go#L397-L409

providers/flagd/pkg/service/service.go
Kavindu-Dodan: Good suggestion. Done - https://github.com/open-feature/go-sdk-contrib/pull/317/commits/3eee6b23362073325def650e01a5973d831bae74

providers/flagd/pkg/configuration.go
Kavindu-Dodan: We do not enforfce gofumpt here. I have run intellij formatter and this is the output it gives 🤔  I think this is simple alphabetic ordering

providers/flagd/pkg/configuration.go
Kavindu-Dodan: not sure this is needed 🤔  I didn't want to add/change the behavior of this logic.

providers/flagd/pkg/configuration.go
thisthat: you are right 👍

providers/flagd/pkg/service/service.go
thisthat: Maybe I am misread, but I think `stream.Receive()` exits only when the stream ends. Hence, if I fail two times to connect, at the third attempt I can connect, after a while the connection drops, the retry starts at 3 instead of 0 🤔

providers/flagd/pkg/configuration.go
thisthat: Agree, this doesn't need to be part of this PR. However, maybe it is worth unifying across languages how to process bool env var. @toddbaert do we have somewhere in the spec a common way for SDKs to handle truthy and falsy values?

providers/flagd/pkg/service/service.go
Kavindu-Dodan: Yup, that's something that the current logic doesn't handle. I will improve this

Kavindu-Dodan: Improvement done with this commit - https://github.com/open-feature/go-sdk-contrib/pull/317/commits/44870f92ea07f9923f2ef3f47acf7ca500aad2c4 

Retry counter and sleep delay now live in its own structure

providers/flagd/pkg/service/service.go
```diff
@@ -432,6 +421,9 @@ func (s *Service) streamClient(ctx context.Context) error {
 	s.logger.V(logger.Info).Info("connected to event stream")
```
thisthat, line 422: Based on the docs (see [Creating Providers](https://github.com/open-feature/flagd/blob/main/docs/other_resources/creating_providers.md#provider-lifecycle-initialization-and-shutdown)), I think here we need to know if we previously failed to emit a `Ready` and `Changed` events

Kavindu-Dodan: The initial ready state is emitted from flagd. This happens at the first event stream connection establishment. 

State change : NOT_READY -> READY   (this event is emitted by SDK, flagd provider do not emit this)

From there on, we can get Config changes, Provider shutdown or connection errors , which are handled by the `streamClient ` logic. 

I do not see the need to store the previous state and could rely on current stream message to produce the valid state

thisthat (approved):
Thanks @Kavindu-Dodan 🙌

toddbaert (approved):

providers/flagd/README.md
beeme1mr, line 1: Please add a "Supported Events" section to the readme. I should look similar to the one available for JS.

https://github.com/open-feature/js-sdk-contrib/tree/main/libs/providers/flagd#supported-events

Kavindu-Dodan: Added with the latest commit. See preview - https://github.com/open-feature/go-sdk-contrib/blob/16fe631890b3ec5ab1eb304b12f9da71628b32ef/providers/flagd/README.md#supported-events

beeme1mr (approved):

PR #327, opened 2023-09-06
feat: implement in-process-flagd provider

bacherfl:
This PR adds the in-process flagd provider, following the [flagd specification](https://github.com/open-feature/flagd/blob/main/docs/other_resources/in-process-providers/specification.md)


odubajDT (approved):
Nice work! :)

providers/in-process-flagd/provider.go
```diff
@@ -0,0 +1,644 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"github.com/open-feature/flagd/core/pkg/eval"
+	"github.com/open-feature/flagd/core/pkg/logger"
+	"github.com/open-feature/flagd/core/pkg/runtime"
+	"github.com/open-feature/flagd/core/pkg/store"
+	"github.com/open-feature/flagd/core/pkg/sync"
```
toddbaert, line 11: really great that we can use these libs. The implementation is surprisingly small.

providers/in-process-flagd/provider.go
```diff
@@ -0,0 +1,644 @@
...
+		log.Fatal("sync provider Init returned error: %w", err)
+	}
+
+	// Start sync provider
+	go provider.startSyncSources(dataSync)
+
+	return provider
+}
+
+func (p *Provider) startSyncSources(dataSync chan sync.DataSync) {
```
toddbaert, line 155:
```suggestion
func (p *Provider) startSyncSource(dataSync chan sync.DataSync) {
```

nit: seems like this is just one source, so perhaps this is a better name?

providers/in-process-flagd/README.md
```diff
@@ -0,0 +1,69 @@
...
+You may also provide additional options to configure the provider client
+```go
+inprocessflagd.WithHost(string)                  // defaults to localhost
+inprocessflagd.WithPort(uint16)                  // defaults to 8013
+inprocessflagd.FromEnv()                         // sets the provider configuration from environment variables
+inprocessflagd.WithSocketPath(string)            // no default, when set a unix socket connection is used (only available for GRPC)
+inprocessflagd.WithTLS(certPath string)          // default of false, if certPath is not given, system certs are used
+inprocessflagd.WithoutCache()                    // disables caching of flag evaluations
+inprocessflagd.WithLRUCache(1000)                // enables LRU caching (see configuring caching section)
+inprocessflagd.WithBasicInMemoryCache()          // enables basic in memory cache (see configuring caching section)
```
toddbaert, line 20: I don't think these caching options are relevant to in-process providers.

providers/in-process-flagd/provider.go
```diff
@@ -0,0 +1,644 @@
...
+		case data := <-dataSync:
+			// resync events are triggered when a delete occurs during flag merges in the store
+			// resync events may trigger further resync events, however for a flag to be deleted from the store
+			// its source must match, preventing the opportunity for resync events to snowball
+			if resyncRequired := p.updateWithNotify(data); resyncRequired {
+				go func() {
+					p.tryReSync(dataSync)
+				}()
+			}
+			if data.Type == sync.ALL {
```
toddbaert, line 178: As of right now, we don't expect any messages other than `sync.ALL`. We may want to log some debug or warning messages if we get others - cc @Kavindu-Dodan

providers/in-process-flagd/provider.go
```diff
@@ -0,0 +1,644 @@
...
+			return nil
+		}
+	}
+}
+
+func (p *Provider) tryReSync(dataSync chan sync.DataSync) {
+	err := p.syncSource.ReSync(p.ctx, dataSync)
+	if err != nil {
+		p.handleConnectionErr(
+			fmt.Errorf("error resyncing sources: %w", err),
```
toddbaert, line 194:
```suggestion
			fmt.Errorf("error resyncing source: %w", err),
```

providers/in-process-flagd/provider.go
```diff
@@ -0,0 +1,644 @@
...
+
+// Status returns the current provider status
+func (p *Provider) Status() of.State {
+	switch p.connectionInfo.state {
+	case stateNotReady:
+		return of.NotReadyState
+	case stateReady:
+		return of.ReadyState
+	case stateStale:
+		// currently there is no STALE state in the of API, so we also return ERROR for STALE
```
toddbaert, line 380: This is a good point... I wonder if we should delay the actual release of this provider until spec `v0.7.0` is implemented in Go (there wouldn't be too much work required).

toddbaert: cc @Kavindu-Dodan @mowies  - we may want to consider something similar for Java/JS, neither of them have `STALE` yet (though there's a [draft PR](https://github.com/open-feature/js-sdk/pull/541) for JS)

Kavindu-Dodan:
@bacherfl I think you forgot to include the `go.mod` file. 

Given this is a new module, we need to configure release please configuration [^1] and set manifest [^2] with initial release version. 

Regarding module name 🤔 shall we prefix this with `flagd` (flagd-in-memory). This should help the folder ordering in provider directory [^3] and give the idea that this is a flagd module.

[^1]: https://github.com/open-feature/go-sdk-contrib/blob/main/release-please-config.json
[^2]: https://github.com/open-feature/go-sdk-contrib/blob/main/.release-please-manifest.json
[^3]: https://github.com/open-feature/go-sdk-contrib/tree/main/providers

providers/in-process-flagd/Makefile
```diff
@@ -0,0 +1,3 @@
+.PHONY: mockgen
+mockgen:
+	mockgen -source=pkg/service/iservice.go -destination=internal/mock/service_mock.go -package=mock
```
Kavindu-Dodan, line 3: 🤔 not sure this was added by mistake, but source doesn't exisit

providers/in-process-flagd/README.md
```diff
@@ -0,0 +1,69 @@
+# In-process Flagd Provider
+
+This provider implements parts of the functionality of [Flagd](https://github.com/open-feature/flagd), and follows its
+specification for [in-process providers](https://github.com/open-feature/flagd/blob/main/docs/other_resources/in-process-providers/specification.md).
+
+## Setup
+To use flagd with the [OpenFeature SDK](https://github.com/open-feature/go-sdk) set the provider to the `openfeature` global singleton as shown below (using default values which align with those of `flagd`)
+```go
+openfeature.SetProvider(inprocessflagd.NewProvider())
```
Kavindu-Dodan, line 9: Naming. we can go for either `inprocessflagd` or `flagdInProcess` 🤔

Any opinion @thisthat @toddbaert  @mowies @odubajDT  ?

providers/in-process-flagd/README.md
Kavindu-Dodan: yeah, this should nearly match with Java provider configurations - https://github.com/open-feature/java-sdk-contrib/tree/main/providers/flagd#configuration-options

providers/in-process-flagd/provider.go
Kavindu-Dodan: yeah, +1 to log other states.

providers/in-process-flagd/provider.go
Kavindu-Dodan: Given that this is part of the spec [^1], state `STALE` should be added to SDK. 

[^1]: https://openfeature.dev/specification/types#provider-events

Kavindu-Dodan: ohh my bad, provider event stale is not yet reflected in provider state. This PR need to be merged for this [^1]

[^1]: https://github.com/open-feature/spec/pull/200

toddbaert: Stale state actually is in the spec `main`, just not release or implemented in very many SDKs: https://github.com/open-feature/spec/blob/main/specification/types.md#provider-status

This will come with spec `0.7.0` release. I'll try to do that tomorrow: https://github.com/open-feature/spec/issues/202

providers/in-process-flagd/README.md
thisthat: I favor the first one: `inprocessflagd`, since it reads better. 
Some reading for guidance on pkg naming: https://go.dev/blog/package-names

providers/in-process-flagd/provider.go
```diff
@@ -0,0 +1,644 @@
...
+	if p.connectionInfo.retries >= p.connectionInfo.maxSyncRetries {
+		p.logger.Error("Number of maximum retry attempts has been exceeded. Going into ERROR state.")
+		p.connectionInfo.state = stateError
+		p.sendProviderEvent(of.Event{
+			EventType: of.ProviderError,
+			ProviderEventDetails: of.ProviderEventDetails{
+				Message: err.Error(),
+			},
+		})
+		return
```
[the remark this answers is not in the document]
thisthat: or we could stack the recovery and the unlock in two defer calls at the top level. 
```go
p.mu.Lock()
defer func() {
  if recoveryFunc != nil {
	<-time.After(p.connectionInfo.backoffDuration)
	recoveryFunc()
  }
}()
defer p.mu.Unlock()
```

Mind the order of defer: https://go.dev/blog/defer-panic-and-recover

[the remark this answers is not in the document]
thisthat: I just realized that the defer option wouldn't work because of the retry logic 😭

providers/in-process-flagd/provider.go
```diff
@@ -0,0 +1,644 @@
...
+
+type ProviderType string
+
+const (
+	defaultMaxSyncRetries      = 5
+	defaultTLS            bool = false
+
+	flagdProxyTLSEnvironmentVariableName          = "FLAGD_PROXY_TLS"
+	flagdProxyCertPathEnvironmentVariableName     = "FLAGD_PROXY_CERT_PATH"
+	flagdMaxSyncRetriesEnvironmentVariableName    = "FLAGD_MAX_SYNC_RETRIES"
```
Kavindu-Dodan, line 29: Regarding max retries 🤔 

For flagd remote evaluations, we can define max retries for sync and go offline when limit is reached. This is because we always try to evaluate flags against remote deployment of flagd. So even if sync is shutoff, evaluation can reach flagd when flagd is back online. 

But for in-process evaluations, our flag definitions comes from remote sync service. If this connection breaks, we have to continuously try to connect to the remote flag source. Otherwise, our store will be stale and without a process restart (application restart) we cannot refresh flags. I think this should be added to flagd specification 

cc - @toddbaert & @thisthat

thisthat: That's a valid point @Kavindu-Dodan 👍

providers/in-process-flagd/internal/logger/logger.go
```diff
@@ -0,0 +1,31 @@
...
+type Logger struct{}
+
+func (l Logger) Init(info logr.RuntimeInfo) {}
+
+func (l Logger) Enabled(level int) bool { return true }
+
+func (l Logger) Info(level int, msg string, keysAndValues ...interface{}) {}
+
+func (l Logger) Error(err error, msg string, keysAndValues ...interface{}) {
+	log.Println("flagd-provider:", err)
```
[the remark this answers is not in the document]
bacherfl: it turns out that right now the internal package is not needed anyway - this was a copy-paste relict from the original flagd provider which this one was based off

providers/in-process-flagd/provider.go
bacherfl: agree @Kavindu-Dodan , that's a good point - I will adapt the implementation to keep retrying in case of an error

bacherfl: if we keep retrying - do you think it still makes sense to eventually go from `STALE` into `ERROR` after a certain amount of retries, or should we get rid of the `MAX_SYNC_RETRIES` env var altogether, and stay in `STALE` for cases where we previously were able to retrieve the flag configs but ran into a interrupted connection afterwards?

Kavindu-Dodan: Good point. For Java provider, I went ahead with status `STALE` and set a max backoff [^1]. If we agree, we can set status to error once we hit the max backoff 🤔 

[^1]: https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/storage/connector/grpc/GrpcStreamConnector.java#L29-L30

bacherfl: Thank you @Kavindu-Dodan - i think the way you implemented that in the Java library makes the most sense here, so I will apply this logic in this provider as well :)

toddbaert: I think if we use the max backoff, it makes sense to make that the "boundary" between STALE and ERROR - but TBH I'm not sure in a server context that anyone would "be okay" with either situation for any substantial amount of time.

I guess a STALE state is tolerable for brief periods and could be characterized as eventual consistency, whereas after max backoff you're at a threshold you've determined to be an error and you might want error-related telemetry to be recorded.

toddbaert (approved):
Approved pending comment [here](https://github.com/open-feature/go-sdk-contrib/pull/327#discussion_r1321695112). I'll add the test suite once this is merged and then we can consider what we'd want before a final release (we may want spec `0.7.0` implemented first).

providers/in-process-flagd/provider.go
bacherfl: Thanks @toddbaert , that makes sense - that's essentially also how the provider is implemented right now. So now this provider will keep retrying with the max backoff being the upper limit, and after exceeding max_retries an error event is sent

providers/in-process-flagd/Makefile
bacherfl: yes, good catch, that was indeed added by accident - removed it

providers/in-process-flagd/provider.go
Kavindu-Dodan: Yes, it makes sense to emit an error if we reach the max back-off limit. I will add this to Java provider

Kavindu-Dodan (approved):

PR #330, opened 2023-09-14
chore: add missing go-cmp dep

toddbaert:
This dep is used in tests but missing in the go.mod


Kavindu-Dodan (approved):

Kavindu-Dodan:
Merging as this is a small dependency fix

PR #331, opened 2023-09-14
chore: add flagd e2e tests

toddbaert:
* adds e2e framework to contribs (tests marked with `// +build e2e` at top of file will run with `make e2e`)
* adds standard `evaluation.feature` gherkin tests for flagd provider

gherkin tests for in-memory provider will come next.

See the tests running in the CI: https://github.com/open-feature/go-sdk-contrib/actions/runs/6188614191/job/16800993762?pr=331#step:8:311

### DIFF
--- a/providers/flagd/e2e/evaluation_test.go
+++ b/providers/flagd/e2e/evaluation_test.go
@@ -1,0 +1,40 @@
+//go:build e2e
+// +build e2e
+
+package e2e
+
+import (
+	"flag"
+	"testing"
+
+	"github.com/cucumber/godog"
+	flagd "github.com/open-feature/go-sdk-contrib/providers/flagd/pkg"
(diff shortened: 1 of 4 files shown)


providers/flagd/e2e/evaluation_test.go
```diff
@@ -0,0 +1,40 @@
+//go:build e2e
+// +build e2e
```
Kavindu-Dodan, line 2: minor - we don't need both and only keep `//go:build e2e`. This is because of the minimum Go version support which is 1.18+

Kavindu-Dodan (approved):

providers/flagd/e2e/evaluation_test.go
toddbaert: Ya my editor (or the go compiler?) automatically adds line 2 when I add line 1. I'm not sure if it's just a VS code thing or not.

PR #334, opened 2023-09-15
fix: error handing, add e2e tests

toddbaert:
* fixes error handling of in-process provider
  * the json-eval returns simple errors (`errors.new(..)`) not ResolutionErrors
* improves the flexibility of the flagd test suite (allows you to supply a provider factory instead of creating a particular provider
* implements the basic `evaluation.feature` test in the in-process provider (which works great!)


providers/flagd-in-process/pkg/provider.go
```diff
@@ -646,3 +624,16 @@ func syncProviderFromConfig(logger *logger.Logger, sourceConfig runtime.SourceCo
 			sourceConfig.Provider, SourceTypeGrpc, SourceTypeKubernetes)
 	}
 }
+
+func mapError(err error) of.ResolutionError {
```
toddbaert, line 628: @bacherfl @Kavindu-Dodan this is the only functional change in this PR - as the description says, the errors thrown in the json-evaluator are just errors, so they need to be handled this way (flagd itself has similar code for mapping them to connect errors).

tests/flagd/pkg/integration/evaluation.go
```diff
@@ -7,12 +7,11 @@ import (
 	"strconv"
 
 	"github.com/cucumber/godog"
-	flagd "github.com/open-feature/go-sdk-contrib/providers/flagd/pkg"
 	"github.com/open-feature/go-sdk/pkg/openfeature"
 )
 
-func InitializeEvaluationScenario(pOptions ...flagd.ProviderOption) func(*godog.ScenarioContext) {
-	providerOptions = pOptions
+func InitializeEvaluationScenario(providerSupplier func() openfeature.FeatureProvider) func(*godog.ScenarioContext) {
```
toddbaert, line 13: The test suites now take a provider factory (so you can use any provider with them).

providers/flagd-in-process/pkg/provider.go
Kavindu-Dodan: Yup, this fully matches with flagd error handling - https://github.com/open-feature/flagd/blob/main/core/pkg/service/flag-evaluation/flag_evaluator.go#L316

Kavindu-Dodan (approved):
changes lgtm

bacherfl (approved):

PR #336, opened 2023-09-19
fix: fractional operator name

toddbaert:
* fixes `fractionalEvaluation` name (should be `fractional`)
* adds e2e suite for custom operators in in-process provider


providers/flagd-in-process/pkg/provider.go
```diff
@@ -127,7 +127,7 @@ func NewProvider(ctx context.Context, opts ...ProviderOption) *Provider {
 		provider.logger,
 		s,
 		eval.WithEvaluator(
-			"fractionalEvaluation",
+			"fractional",
```
toddbaert, line 130: This is the only functional change.

beeme1mr (approved):

.github/component_owners.yml
```diff
@@ -10,6 +10,10 @@ components:
   providers/flagd:
     - Kavindu-Dodan
     - toddbaert
+  providers/flagd-in-process:
+    - bacherfl
+    - Kavindu-Dodan
+    - toddbaert
```
Kavindu-Dodan, line 16: 🎉

providers/flagd-in-process/pkg/provider.go
Kavindu-Dodan: Will this be a breaking change 🤔  ?

Kavindu-Dodan (approved):

providers/flagd-in-process/pkg/provider.go
toddbaert: sort-of... but it was a bug in the first place. As per spec/doc the operator was always supposed to be called `fractional`.

Also we haven't released it yet :smile:

toddbaert:
@beeme1mr @Kavindu-Dodan note, I've added this package to to release please config. I think we can release after this.

providers/flagd-in-process/pkg/provider.go
Kavindu-Dodan: Ahh yeah, sorry I forgot the fact that the Go in-process provider is a separate package :)

PR #338, opened 2023-09-28
feat: Add Unleash Provider

liran2000:
Add [Unleash](https://www.getunleash.io/) Provider.

See [readme](https://github.com/liran2000/go-sdk-contrib/blob/55ec211520c75761b34abafee762bbbb1be8b856/providers/unleash/README.md#unofficial-unleash-openfeature-go-provider) for details.

@sighphyre @gastonfournier @ivarconr you are welcome to review from Unleash perspective.

@Kavindu-Dodan @toddbaert you are welcome to review from OpenFeature perspective.


ivarconr:
nice 👍🏼

providers/unleash/pkg/provider.go
```diff
@@ -0,0 +1,168 @@
...
+	"github.com/Unleash/unleash-client-go/v3/api"
+	of "github.com/open-feature/go-sdk/pkg/openfeature"
+)
+
+const providerNotReady = "Provider not ready"
+const generalError = "general error"
+
+type Provider struct {
+	status      of.State
+	eventStream chan of.Event
```
Kavindu-Dodan, line 17: Given that there's no event stream usage, this can be safely removed (including `EventChannel` implementation)

From OF point of view, it is not essential for a provider to support events :)

providers/unleash/pkg/provider.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	provider := &Provider{
+		status: of.ReadyState,
+	}
+
+	return provider, nil
+}
+
+func (p *Provider) Status() of.State {
```
Kavindu-Dodan, line 38: We can complete the `StateHandler` interface implementation by implementing `Init(evaluationContext EvaluationContext) error`

And the initialization can wrap the `unleash.Initialize(...)` , including handling errors.

Kavindu-Dodan (approved):
Overl looks good from OF point of view. Please consider the review commits and related implementations. 

@toddbaert hope you can have a look as well

providers/unleash/README.md
```diff
@@ -0,0 +1,60 @@
...
+	}
+
+	provider, err := unleashProvider.NewProvider(providerConfig)
+
+	ctx := context.Background()
+
+	resolution := provider.BooleanEvaluation(ctx, "variant-flag", false, nil)
+	enabled := resolution.Value
+
+    resolution := provider.StringEvaluation(ctx, "variant-flag", "", nil)
```
sighphyre, line 46:
```suggestion
	resolution := provider.StringEvaluation(ctx, "variant-flag", "", nil)
```

providers/unleash/pkg/provider.go
```diff
@@ -0,0 +1,168 @@
...
+	eventStream chan of.Event
+}
+
+func NewProvider(providerConfig ProviderConfig) (*Provider, error) {
+	return NewProviderWithContext(context.Background(), providerConfig)
+}
+
+func NewProviderWithContext(ctx context.Context, providerConfig ProviderConfig) (*Provider, error) {
+	if providerConfig.Options != nil {
+		unleash.Initialize(
```
sighphyre, line 26: I think the lint is correct here. Initialize can fail, some of these are hard error cases like a broken URL, so we probably need to handle this somehow

providers/unleash/pkg/provider.go
```diff
@@ -0,0 +1,168 @@
...
+				Value: defaultValue,
+				ProviderResolutionDetail: of.ProviderResolutionDetail{
+					ResolutionError: of.NewGeneralResolutionError(generalError),
+					Reason:          of.ErrorReason,
+				},
+			}
+		}
+	}
+
+	feature := unleash.GetVariant(flag)
```
sighphyre, line 147: From a purely Unleash perspective, this is a variant, not a feature. If it makes more sense within this context to use `feature`, that's totally fine, but renaming this variable to variant makes a lot more sense to my eyes

sighphyre, line 147: I don't think we're passing the Unleash context here, which is definitely needed for some toggles to successfully evaluate

providers/unleash/pkg/provider.go
```diff
@@ -0,0 +1,168 @@
...
+				Value: defaultValue,
+				ProviderResolutionDetail: of.ProviderResolutionDetail{
+					ResolutionError: of.NewGeneralResolutionError(generalError),
+					Reason:          of.ErrorReason,
+				},
+			}
+		}
+	}
+
+	res := unleash.IsEnabled(flag, unleash.WithFallback(defaultValue))
```
sighphyre, line 84: Same as below, I think this needs to be passed the evalCtx parameter

providers/unleash/pkg/provider_test.go
```diff
@@ -0,0 +1,99 @@
+package unleash_test
```
sighphyre, line 1: I'd really like to see some tests that check that the context passed in is respected. In the test data you have a "users-flag", which would need a context with userid on it to successfully return true. That might be a nice flag to build some tests off of for this

toddbaert:
I think you might have force pushed something in error, @liran2000, I'm not seeing the changed files referenced so far in the review.

I'll check back again later.

providers/unleash/pkg/provider.go
```diff
@@ -0,0 +1,236 @@
...
+		providerConfig: providerConfig,
+	}
+	return provider, nil
+}
+
+func (p *Provider) Init(evaluationContext of.EvaluationContext) error {
+	err := unleash.Initialize(
+		p.providerConfig.Options...,
+	)
+	p.status = of.ReadyState
```
toddbaert, line 33: I think we'd want to set the state to error if `err == nil`.

providers/unleash/pkg/provider.go
liran2000: Thanks, removed it.
Unleash [supports events listener](https://github.com/Unleash/unleash-client-go/blob/2b58cc4de44e179461a22a29660f049d24d50d4c/unleash_test.go#L18), if you want to take it for a separate PR, by something like converting the unleash events to OpenFeature events, we can discuss separately.

providers/unleash/pkg/provider.go
```diff
@@ -0,0 +1,236 @@
...
+		return &unleashContext.Context{}, nil
+	}
+
+	unleashContext := &unleashContext.Context{}
+
+	custom := make(map[string]string)
+	for key, origVal := range evalCtx {
+		val, ok := toStr(origVal)
+		if !ok {
+			return nil, fmt.Errorf("key `%s` can not be converted to string", key)
```
toddbaert, line 198: May want to use "Invalid Context" error here: https://github.com/open-feature/go-sdk/blob/d8da306f06a5159abdb088b91e78fb59cb0b68a5/pkg/openfeature/resolution_error.go#L19

providers/unleash/pkg/provider_test.go
liran2000: [Added](https://github.com/open-feature/go-sdk-contrib/pull/338/files#diff-1cb45209a551a2fbaa8ebc0e5d4be5ae1f5e3381048e165e0ff15c22034b21c4R143).

toddbaert (approved):
LGTM, but I will await approval from one Unleashed contributor before merge!

liran2000:
@sighphyre @Kavindu-Dodan @toddbaert thank you for the valuable feedback, I addressed your comments.

* had to rebase due to missing _sign-off_ in some commits from a different IDE.

.release-please-manifest.json
```diff
@@ -8,5 +8,6 @@
     "providers/go-feature-flag": "0.1.29",
     "providers/flagsmith": "0.1.4",
     "providers/launchdarkly": "0.1.3",
-    "tests/flagd": "1.3.1"
+    "tests/flagd": "1.3.1",
+    "tests/unleash": "0.0.1-alpha"
```
toddbaert, line 12: @liran2000 I guess this was a copy paste error, and you mean to do `providers/unleash`. I will force a quick fix for this, right now this is causing a 1.0 release which I don't think you want.

Sorry I missed this! :sweat_smile:

toddbaert: Fixed and release looks good now: https://github.com/open-feature/go-sdk-contrib/pull/346

PR #348, opened 2023-10-12
feat: Add Harness provider

liran2000:
Add [Harness](https://developer.harness.io/docs/feature-flags) GO Provider.

@davejohnston you are welcome to review from Harness point of view.

@Kavindu-Dodan @toddbaert you are welcome to review from OpenFeature perspective.


davejohnston (commented):
Hi @liran2000 thanks for putting this together.  At a high level it looks good.
I'm just going to take one more look over it and then I'll approve.  

We recently made a change in the goland SDK around initialisation which might help determine when the SDK is ready or not (once its connected and fetched all its config).

providers/harness/README.md
```diff
@@ -0,0 +1,65 @@
...
+    t.Fail()
+}
+err = provider.Init(of.EvaluationContext{})
+if err != nil {
+    t.Fail()
+}
+
+ctx := context.Background()
+
+target := map[string]interface{}{
```
toddbaert, line 51:
```suggestion
context := map[string]interface{}{
```

"target" makes sense, but I think `context` is probably best especially for documentation... unless "target" is a word Harness uses specifically, then I think this is fine.

providers/harness/README.md
```diff
@@ -0,0 +1,65 @@
...
+    "Identifier": "john",
+    "Firstname":  "John",
+    "Lastname":   "Doe",
+    "Email":      "john@doe.com",
+}
+
+resolution := provider.BooleanEvaluation(ctx, "TestTrueOn", false, target)
+if resolution.Value != true {
+    t.Fatalf("Expected one of the variant payloads")
+}
```
toddbaert, line 61: I think you probably want to use the actual SDK here, not the provider, or people might get confused.

providers/harness/pkg/provider.go
```diff
@@ -0,0 +1,310 @@
...
+	"github.com/harness/ff-golang-server-sdk/types"
+	of "github.com/open-feature/go-sdk/pkg/openfeature"
+)
+
+const providerNotReady = "Provider not ready"
+const generalError = "general error"
+
+type Provider struct {
+	providerConfig ProviderConfig
+	harnessClient  *client.CfClient
```
toddbaert, line 19: @davejohnston is there any risk having multiple clients in a runtime? Do they share any global state? Some SDKs don't play well with that. I suspect not, just pointing it out.

providers/harness/pkg/provider.go
```diff
@@ -0,0 +1,310 @@
...
+
+func (p *Provider) Init(evaluationContext of.EvaluationContext) error {
+	harnessClient, err := harness.NewCfClient(p.providerConfig.SdkKey, p.providerConfig.Options...)
+	if err != nil {
+		p.status = of.ErrorState
+	} else {
+		p.status = of.ReadyState
+		p.harnessClient = harnessClient
+	}
+	return err
```
toddbaert, line 39: Does harness include any sort of change events? We don't need it for first release, but we could maybe make an issue for it.

providers/harness/pkg/provider.go
```diff
@@ -0,0 +1,310 @@
...
+			},
+		}
+	}
+
+	res, _ := p.harnessClient.BoolVariation(flag, harnessTarget, defaultValue)
+	return of.BoolResolutionDetail{
+		Value:                    res,
+		ProviderResolutionDetail: of.ProviderResolutionDetail{},
+	}
+}
```
toddbaert, line 100: You might be able to use generics to reduce some duplication in these functions - but I'm not very experienced with Go generics, so I could be wrong.

providers/harness/README.md
toddbaert: Nevermind! I can see that "target" is Harness' term.

providers/harness/pkg/provider.go
```diff
@@ -0,0 +1,310 @@
...
+		if !ok {
+			return nil, fmt.Errorf("key `%s` can not be converted to string", key)
+		}
+
+		switch key {
+		case "Identifier":
+			harnessTarget.Identifier = val
+		case "Name":
+			harnessTarget.Name = val
+		default:
```
toddbaert, line 288: if `Name` or `Identifier` are special fields for identifying a target/user, we might want to map the `targetingKey` into one of these, and document which it is. We also may want to provide an option for which it's mapped to, if both are used for that purpose.

providers/harness/pkg/provider.go
liran2000: I didn't see events in Harness GO SDK which can be mapped to provider events. If so then it can be address as a follow-up issue yes.

providers/harness/pkg/provider.go
davejohnston: @toddbaert no we should be good here.  Not all our SDKs behave well but golang is one that we have initialised multiple instances in a single application.

providers/harness/pkg/provider.go
davejohnston: @liran2000 I think this is something we could add into the golang SDK.  Currently we expose a func to tell you if it is initialised.  We debated about exposing a channel to send the event on, but opted not to.  I think what we can do is make it a option to pass into the constructor, where you can pass you're own channel in and get a single when we successfully init

providers/harness/pkg/provider.go
liran2000: I re-arranged it a bit, such that each evaluation is now more minimal.

liran2000:
> Hi @liran2000 thanks for putting this together. At a high level it looks good. I'm just going to take one more look over it and then I'll approve.

Thanks @davejohnston, I made some changes according to comments, you can have a look and write here if you approve.

providers/harness/pkg/provider.go
liran2000: @toddbaert trying to understand, why would this can cause having multiple clients in a runtime? can you please elaborate ?

toddbaert: Some vendors assume only one SDK client (not openfeature client) exists in the runtime. They share some global state in some cases.

According to @davejohnston this isn't an issue here, at least with Harness' golang SDK.

providers/harness/pkg/provider.go
toddbaert: I think that would be a cool feature. Personally I've always found change events in SDKs useful.

providers/harness/README.md
```diff
@@ -0,0 +1,74 @@
...
+* Provider Object evaluation gets Harness JSON evaluation.
+* Other provider types evaluation gets Harness matching type evaluation.
+
+## Usage
+Harness OpenFeature Provider is using Harness GO SDK.
+
+### Evaluation Context
+Evaluation Context is mapped to Harness [target](https://developer.harness.io/docs/feature-flags/ff-sdks/server-sdks/feature-flag-sdks-go-application/#add-a-target).
+OpenFeature targetingKey is mapped to _Identifier_, _Name_ is mapped to _Name_ and other fields are mapped to Attributes 
+fields.
```
toddbaert, line 24: Thanks!

toddbaert (approved):
LGTM, will merge pending @davejohnston 's approval.

davejohnston (approved):
Looks good, thanks for you’re contribution

Kavindu-Dodan (approved):

PR #373, opened 2023-11-04
feat: upgrade to unleash-client-go v3.9.0

liran2000:
- upgrade to unleash-client-go v3.9.0
- refactor tests provider init

due to #368 issue.
@toddbaert @sighphyre @Kavindu-Dodan


providers/unleash/pkg/provider_config.go
```diff
@@ -4,10 +4,7 @@ import (
 	"github.com/Unleash/unleash-client-go/v3"
 )
 
-// ProviderOptions is the struct containing the provider options you can
-// use while initializing GO Feature Flag.
-// To have a valid configuration you need to have an Endpoint or GOFeatureFlagConfig set.
+// ProviderOptions is the struct containing the provider options.
 type ProviderConfig struct {
-	// Endpoint contains the DNS of your GO Feature Flag relay proxy (ex: http://localhost:1031)
```
toddbaert, line 11: oops! I guess I missed this too, first time.

providers/unleash/pkg/provider_test.go
```diff
@@ -164,8 +83,52 @@ func TestBooleanEvaluationByUser(t *testing.T) {
...
 		t.Fatalf("Expected feature to be enabled")
 	}
 }
+
+// global cleanup
+func cleanup() {
+	provider.Shutdown()
+}
+
+func TestMain(m *testing.M) {
```
toddbaert, line 97: I didn't know about `TestMain`!

toddbaert (approved):
I'm not sure this is a `feat`. Since there is already a pending release for this artifact, you could call this a `chore` if you want, which is probably most accurate, and still include it in the next release. I'll leave that up to you though, in this case I don't mind so much since it's still an early version.

toddbaert (approved):

PR #394, opened 2023-11-23
feat: consolidate flagd in-process provider

### LINKED ISSUE #393: [flagd] Consolidate flagd provider sdks
**Background**

flagd go in-process provider [^1] is maintained in its own module. This does not align with other languages (ex:- Java, js) where in-process mode is enabled through the same sdk. This creates more confusion and adds maintenance complexity. 

**Action**

Deprecate the dedicated in-process provider implementation and consolidate the implementation into the old flagd go sdk.

- [x] Consolidate in-process provider implementation to the old flagd go sdk
- [x] Deprecate the dedicated in-process provider and release the new go sdk provider with built in in-process support
- [ ] Remove the dedicated provider implementation 

[^1]:  https://github.com/open-feature/go-sdk-contrib/tree/main/providers/flagd-in-process

Kavindu-Dodan:
## This PR

Consolidate flagd in-process provider to existing flagd provider 

This change brings in-process provider to the exisiting flagd provider. 
```go
provider := flagd.NewProvider(
        flagd.WithInProcessResolver(),
        flagd.WithHost("localhost"),
        flagd.WithPort(8013))
openfeature.SetProvider(provider)
```

The change set contains minor packaging refactoring to facilitate new in-process packaging 

### Related Issues

Related to #393

### Follow-up Tasks

Deprecate dedicated in-process provider package after the initial release


providers/flagd/pkg/service/in_process/service.go
```diff
@@ -0,0 +1,288 @@
...
+}
+
+func (i *InProcess) EventChannel() <-chan of.Event {
+	return i.events
+}
+
+// mapError is a helper to map evaluation errors to OF errors
+func mapError(err error) of.ResolutionError {
+	switch err.Error() {
+	case model.FlagNotFoundErrorCode, model.FlagDisabledErrorCode:
```
toddbaert, line 279: Flag disabled shouldn't be an error, I think. It's a reason, and the deftault should be returned without an error being indicated.

See reasons in the resolution details: https://openfeature.dev/specification/types#resolution-details, and the [FAQ](https://openfeature.dev/docs/reference/concepts/provider#why-isnt-there-an-error-code-for-disabled-flags) on this situation.

Kavindu-Dodan: hmm, I think we need to revisit this issue. We have different handling of disabled state.

- Flagd core : generates an error [^1]
- Java provider : consider this as an error [^2]
- JS provider : consider this as NOT an error [^3]

And the go provider currently consider this as an error - same for current in-process go provider 

[^1]: https://github.com/open-feature/flagd/blob/main/core/pkg/service/flag-evaluation/flag_evaluator.go#L318-L319
[^2]: https://github.com/open-feature/java-sdk-contrib/blob/main/providers/flagd/src/main/java/dev/openfeature/contrib/providers/flagd/resolver/process/InProcessResolver.java#L159-L162
[^3]: https://github.com/open-feature/js-sdk-contrib/blob/main/libs/shared/flagd-core/src/lib/flagd-core.ts#L85-L91

toddbaert: Ok, ya, we should be consistent here. The reasoning is that disabled flags aren't really an exceptional circumstance (it's a supported configuration in most systems) therefore it shouldn't involve an exception-based control flow.

But anyway, I agree it seems to be a bigger issue than just this PR, so do what you can here and I won't consider it a blocker. We can open other issues where it's relevant.

Kavindu-Dodan:
> There's some integration tests [here](https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/e2e/evaluation_test.go). I think you'll have to modify this to run twice, once with the RPC and once with the in-process provider. The `ScenarioInitializer` takes a function which returns a provider instance, so you can return the instance you need for each run. You will also need to run the json_eval test, which currently runs in the in-process provider only(see [here](https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd-in-process/e2e/json_evalutor_test.go)).
> 
> That will establish the provider works as expected when configured for in-process. The Java and JS repos do similar things to tests both resolvers.

Done, added e2e tests for InProcess provider. Also added missing json evaluator test

toddbaert:
> > There's some integration tests [here](https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/e2e/evaluation_test.go). I think you'll have to modify this to run twice, once with the RPC and once with the in-process provider. The `ScenarioInitializer` takes a function which returns a provider instance, so you can return the instance you need for each run. You will also need to run the json_eval test, which currently runs in the in-process provider only(see [here](https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd-in-process/e2e/json_evalutor_test.go)).
> > That will establish the provider works as expected when configured for in-process. The Java and JS repos do similar things to tests both resolvers.
> 
> Done, added e2e tests for InProcess provider. Also added missing json evaluator test

Thanks! I will re-review.

Just for clarity, I don't think the json test was really "missing" before, because it specifically tests a bunch of logic only the in-process provider does (for example, sem_ver), so it wasn't relevant here before. I've tried to build these suites so they cover different sets of functionality:

`evaluation.feature`: general evaluation, useful for every provider
`flagd.feature`: flagd-specific stuff, but applicable to both rpc and in-process providers (zero-value handling, flag change events, etc)
`flagd-json-evaluator.feature`: all the specific custom json-logic ops, not relevenat for flagd-rpc providers, since flagd itself does this work

Despite the fact that the rpc provider doesn't really need to, running all the test for both resolver configurations is fine; they are very quick.

toddbaert (approved):

PR #396, opened 2023-12-01
feat: deprecate dedicated flagd in-process provider

Kavindu-Dodan:
## This PR

Comples part of #393

Adds deprecation notice on flagd in-process provider package.

### DIFF
--- a/providers/flagd-in-process/pkg/provider.go
+++ b/providers/flagd-in-process/pkg/provider.go
@@ -59,6 +59,7 @@ type connectionInfo struct {
 
 type connectionState int
 
+// DEPRECATED : Please use flagd with WithInProcessResolver option instead of this dedicated provider
 type Provider struct {
 	ctx                   context.Context
 	cancelFunc            context.CancelFunc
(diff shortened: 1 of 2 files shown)


providers/flagd-in-process/pkg/provider.go
```diff
@@ -59,6 +59,7 @@ type connectionInfo struct {
 
 type connectionState int
 
+// DEPRECATED : Please use flagd with WithInProcessResolver option instead of this dedicated provider
```
thisthat, line 62: It should be like the following:
```suggestion
// Deprecated: Please use flagd with WithInProcessResolver option instead of this dedicated provider
```
You can see more info around this here: https://github.com/golang/go/issues/10909

providers/flagd-in-process/pkg/provider.go
```diff
@@ -80,6 +81,7 @@ type ProviderConfiguration struct {
 
 type ProviderOption func(*Provider)
 
+// DEPRECATED : Please use flagd with WithInProcessResolver option instead of this dedicated provider
```
Kavindu-Dodan, line 84:
```suggestion
// Deprecated : Please use flagd with WithInProcessResolver option instead of this dedicated provider
```

thisthat (approved):

PR #408, opened 2023-12-13
fix: nil client causes panic

### LINKED ISSUE #407: [bug] Panic in flagd provider
The flagd provider sometimes panics on startup. This is because[ the client is `nil` until init is called](https://github.com/open-feature/go-sdk-contrib/blob/febd9b89878f2a998df39698e752dd3583ffd56e/providers/flagd/pkg/service/rpc/service.go#L77), which happens in a new goroutine and therefore is sometimes not ready.

Related to: https://github.com/open-feature/go-sdk/issues/238

thisthat:
## This PR

- fixes #407 

related to: https://github.com/open-feature/go-sdk/issues/238

### DIFF
--- a/providers/flagd/pkg/provider.go
+++ b/providers/flagd/pkg/provider.go
@@ -3,13 +3,13 @@ package flagd
 import (
 	"context"
 	"fmt"
-
 	"github.com/go-logr/logr"
 	"github.com/open-feature/go-sdk-contrib/providers/flagd/internal/cache"
 	"github.com/open-feature/go-sdk-contrib/providers/flagd/internal/logger"
 	"github.com/open-feature/go-sdk-contrib/providers/flagd/pkg/service/in_process"
 	rpcService "github.com/open-feature/go-sdk-contrib/providers/flagd/pkg/service/rpc"
 	of "github.com/open-feature/go-sdk/openfeature"
+	"sync"
 )
 
 type Provider struct {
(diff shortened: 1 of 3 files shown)


providers/flagd/pkg/provider.go
```diff
@@ -158,6 +161,12 @@ func (p *Provider) ObjectEvaluation(
 	return p.service.ResolveObject(ctx, flagKey, defaultValue, evalCtx)
 }
 
+func (p *Provider) setStatus(status of.State) {
+	p.mtx.Lock()
+	defer p.mtx.Unlock()
+	p.status = status
```
toddbaert, line 167: I would assume this is atomic and therefore locking doesn't help? :thinking:

toddbaert: Though I think you want to make sure this can run DURING init.

I think I thread isn't blocked by a lock it already owns, right? So I guess you might want to lock in init() and well as here.

bacherfl: I think we do need it, yes - if you execute the tests with the `-race` flag enabled, a data race error occurs, due to `p.status` being set in the goroutine created within the `Init()` function, while being accessed via the `Status()` function in the test

toddbaert (approved):

toddbaert:
Thanks @bacherfl and @thisthat !

Issue #417, opened 2024-01-09
[flagd] validate race conditions of in-process provider


Kavindu-Dodan:
As suggested by [this](https://github.com/open-feature/java-sdk-contrib/pull/614#discussion_r1445084437) comment, our concurrency logic needs to be validated for race conditions. Though we expect initialization to be called once, there could be edge cases such as multiple provider registrations which could invoke unwanted initialization attempts.

**AC**

In-process provider implementation is validated for potential race conditions and fixed.

PR #418, opened 2024-01-10
feat: flagd - make sure provider initialize only once

Kavindu-Dodan:
## This PR

Fixed https://github.com/open-feature/go-sdk-contrib/issues/417 by adding locks and validations to avoid provider reinitializations. This is done at provider level hence no changes necessary at service level. 

Added a test to validate the logic.

### DIFF
--- a/providers/flagd/pkg/provider.go
+++ b/providers/flagd/pkg/provider.go
@@ -13,6 +13,7 @@ import (
 )
 
 type Provider struct {
+	initialized           bool
 	logger                logr.Logger
 	providerConfiguration *providerConfiguration
 	service               IService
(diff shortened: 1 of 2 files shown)


providers/flagd/pkg/provider.go
```diff
@@ -13,6 +13,7 @@ import (
 )
 
 type Provider struct {
+	initialized           bool
```
thisthat, line 16: maybe we can be more explicit and pass a false value into the `NewProvider` func

thisthat (approved):
LGTM, I left only a minor suggestion :)

providers/flagd/pkg/provider.go
Kavindu-Dodan: We don't have to pass the value for `initialized` state through `NewProvider`as it's not up to the end user to set it.
But I explicitly assigned a `false` at struct initialization time.

thisthat: Sorry, I wrote it poorly. What you've done is exactly what I was suggesting :)

toddbaert (approved):

PR #442, opened 2024-01-31
fix: improvements to the shutdown of provider & support go 1.20

Kavindu-Dodan:
## This PR

- Fix the shutdown of the flagd rpc provider and improves shutdown trigger of the in-process provider 
- Downgrade to go version 1.20 as this cover two most recent go versions (1.20 & 1.21)

Depends on https://github.com/open-feature/go-sdk-contrib/pull/443 ✅


providers/flagd/pkg/service/in_process/service.go
```diff
@@ -90,11 +90,8 @@ func (i *InProcess) Init() error {
 		return err
 	}
 
+	initOnce := parallel.Once{}
 	syncInitSuccess := make(chan interface{})
-	readyOnce := parallel.OnceFunc(func() {
```
Kavindu-Dodan, line 94: `OnceFunc` is from Go 1.21, making the provider incompatible with older go versions

providers/flagd/go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/open-feature/go-sdk-contrib/providers/flagd
 
-go 1.21
```
toddbaert, line 3: Do you have any idea why this keeps being added?

providers/flagd/pkg/service/in_process/service.go
toddbaert: bummer, I like that one, but I guess it's just syntax sugar.

toddbaert (approved):
Approved but 2 optional suggestions:

- change the title to specifically say what was fixed if you can (what negative behavior is fixed)
- add tests if possible, though it might be tough. If thees fixes are too hard to test that's understandable

providers/flagd/go.mod
Kavindu-Dodan: This comes from Renovate and there's an option to control the version - https://docs.renovatebot.com/golang/#go-binary-version 

However, given this is a mono repo with multiple providers, we cannot enforce this as each provider may support their desired go version. 

As maintainers however, we should allow our CI to run with the latest version of Go as there's a backward compatibility promise.

providers/flagd/pkg/service/in_process/service.go
Kavindu-Dodan: Agree, it's straightforward but breaks when used with a lower go version

PR #443, opened 2024-01-31
feat: domain specific provider for e2e tests

Kavindu-Dodan:
## This PR

Improves scenario initialization by using domain-specific providers and cliens

### DIFF
--- a/tests/flagd/pkg/integration/integration.go
+++ b/tests/flagd/pkg/integration/integration.go
@@ -15,11 +15,12 @@ type ctxStorageKey struct{}
 
 // ctxClientKey is the key used to pass the openfeature client across context.Context
 type ctxClientKey struct{}
+var domain = "flagd-e2e-tests"
 
 func aFlagdProviderIsSet(ctx context.Context) (context.Context, error) {
 	readyChan := make(chan struct{})
 
-	err := openfeature.SetProvider(test_provider_supplier())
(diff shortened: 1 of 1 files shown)


beeme1mr (approved):

tests/flagd/pkg/integration/integration.go
```diff
@@ -15,11 +15,12 @@ type ctxStorageKey struct{}
 
 // ctxClientKey is the key used to pass the openfeature client across context.Context
 type ctxClientKey struct{}
+var domain = "flagd-e2e-tests"
```
toddbaert, line 18: This will still use the same domain/provider between runs, but it's an improvement.

You could attempt to inject a different domain each run, either as a param somehow or through the Go context, maybe?

toddbaert (approved):
Approved, but consider: https://github.com/open-feature/go-sdk-contrib/pull/443/files#r1473380263

tests/flagd/pkg/integration/integration.go
Kavindu-Dodan: unfortunately, this is not straightforward. We register a provider supplier, which is a one-time operation (supplier yields a new provider ). And there's no way to customize this unless we do this per test invoke, which means a change in gherkin test defintion 🤔

Kavindu-Dodan: I will resolve this conversation with above jutification :)

PR #445, opened 2024-02-03
feat: Add Statsig provider

liran2000:
Add [Statsig](https://statsig.com/) GO Provider.

@MovieStoreGuy, @jasonwzm, @xinlili-statsig, @kenny-statsig, @Kavindu-Dodan, @toddbaert you are welcome to review.


providers/statsig/go.mod
```diff
@@ -0,0 +1,19 @@
+module github.com/open-feature/go-sdk-contrib/providers/statsig
+
+go 1.19
```
Kavindu-Dodan, line 3: You could consider upgrading to 1.21 . Go 1.22 is already out and soon we plan to upgrade go-sdk to 1.20 (and 1.21 eventually)

providers/statsig/pkg/provider.go
```diff
@@ -0,0 +1,320 @@
...
+		flagMetadata["LogExposure"] = config.LogExposure
+		flagMetadata["Name"] = config.Name
+		flagMetadata["RuleID"] = config.RuleID
+		return of.InterfaceResolutionDetail{
+			Value: config.Value[flag],
+			ProviderResolutionDetail: of.ProviderResolutionDetail{
+				FlagMetadata: flagMetadata,
+			},
+		}
+	} else {
```
Kavindu-Dodan, line 182: I think an explicit check for `LAYER` is better here

Kavindu-Dodan (commented):
Except for minor remarks, this looks good from OF poitn of view

providers/statsig/go.mod
liran2000: done

providers/statsig/pkg/provider.go
liran2000: done

providers/statsig/pkg/provider.go
```diff
@@ -0,0 +1,328 @@
...
+	return of.BoolResolutionDetail{
+		Value:                    res,
+		ProviderResolutionDetail: of.ProviderResolutionDetail{},
+	}
+}
+
+func (p *Provider) FloatEvaluation(ctx context.Context, flag string, defaultValue float64, evalCtx of.FlattenedContext) of.FloatResolutionDetail {
+	res := p.ObjectEvaluation(ctx, flag, defaultValue, evalCtx)
+	return of.FloatResolutionDetail{
+		Value:                    res.Value.(float64),
```
kenny-statsig, line 99: If it fails type assertion, should it return the default value? Currently, it could silently error

kenny-statsig (approved):
Looks good! Thank you

providers/statsig/pkg/provider.go
liran2000: updated

providers/statsig/README.md
```diff
@@ -0,0 +1,81 @@
...
+## Usage
+Statsig OpenFeature Provider is using Statsig GO SDK.
+
+## Usage Example
+
+```go
+import (
+  statsigProvider "github.com/open-feature/go-sdk-contrib/providers/statsig/pkg"
+	of "github.com/open-feature/go-sdk/openfeature"
+	statsig "github.com/statsig-io/go-sdk"
```
Kavindu-Dodan, line 34: nit : format

```suggestion
import (
  statsigProvider "github.com/open-feature/go-sdk-contrib/providers/statsig/pkg"
  of "github.com/open-feature/go-sdk/openfeature"
  statsig "github.com/statsig-io/go-sdk"
```

providers/statsig/pkg/provider.go
```diff
@@ -0,0 +1,436 @@
...
+	case float32, float64:
+		return fmt.Sprintf("%.6f", v), true
+	case bool:
+		return fmt.Sprintf("%t", v), true
+	default:
+		return "", false
+	}
+}
+
+type FeatureConfigType string
```
Kavindu-Dodan, line 413: Seems types defined beyond this point are internal only, so use the lowercase first letter to avoid exporting

Kavindu-Dodan (approved):
I don't see any blockers other than for the comments I left

providers/statsig/pkg/provider.go
liran2000: Using it at tests, and similarly the consumer can use it, no ?

Kavindu-Dodan: Noticed that you use it in the example, where it's a part of the context 👍

PR #452, opened 2024-02-13
chore: bump Go to version 1.21

odubajDT:
Update contrib modules to Go 1.21

PR #455, opened 2024-02-27
chore: fix tests

toddbaert:
Something about this dependencies in flagd causes the unleash provider to break. :thinking:


toddbaert:
Looks like this is fixed by https://github.com/open-feature/go-sdk-contrib/pull/452

PR #464, opened 2024-02-27
chore: fix initialization race conditions in tests

Kavindu-Dodan:
## This PR

Fixes initialization race conditions which were observed in tests

@liran2000  could you please check whether you need manul provider initialization in the test [1]. This collides with OF provider registration

[1] - https://github.com/open-feature/go-sdk-contrib/blob/main/providers/unleash/pkg/provider_test.go#L121-L123

### DIFF
--- a/providers/unleash/pkg/provider.go
+++ b/providers/unleash/pkg/provider.go
@@ -3,6 +3,7 @@ package unleash
 import (
 	"context"
 	"fmt"
+	"sync"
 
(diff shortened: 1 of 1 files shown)


liran2000:
@Kavindu-Dodan created #465.

Kavindu-Dodan:
> @Kavindu-Dodan created #465.

Thank you, I will close this PR and accept the fix proposed through https://github.com/open-feature/go-sdk-contrib/pull/465

PR #477, opened 2024-03-15
feat: introduce OFREP provider  

### LINKED ISSUE #473: [OFREP] create the initial version of the remote evaluation protocl
**Background**

With the merging of https://github.com/open-feature/protocol/pull/2, we have the first version of OFREP ready. And there are already efforts ongoing to implement flag management system support for the OFREP [1]

**Goal**

As a contributor and maintainer of the OF Go SDK, I would like to have the first version of the OFREP provider in Go SDK. 

For the first implementation, consider the following,

- [ ] Only rely on single flag evaluations as Go SDK works with the dynamic context paradigm
- [ ] Use well-known, well-established libraries where needed
- [ ] No caching in the initial version (keep it simple) 
- [ ] Include a documentation and a usage guide

**AC**

The first version of the OFREP Go SDK is ready

[1] - https://github.com/open-feature/flagd/issues/1245

Kavindu-Dodan:
## This PR

Fixes #473 and introduce the Go OFREP provider 🎉 

### Usage

Consider the following code snippet for basic usage,
```go
provider := ofrep.NewProvider(
                     "http://localhost:8016",
                      ofrep.WithBearerToken("TOKEN"))
```

### Implementation

This implementation relies on OFREP single flag evaluation endpoint. And there is no caching layer at the current implementation and flag evaluations always rely on remote evaluation results.

I have introduced modules where appropriate which should allow us to expand the implementation with a caching layer and other OFREP extensions that we plan to introduce

### DIFF
--- a/providers/ofrep/internal/outbound/http.go
+++ b/providers/ofrep/internal/outbound/http.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"io"
 	"net/http"
 	"net/url"
 	"time"
@@ -20,6 +21,12 @@ type Configuration struct {
 	BaseURI   string
 	Callbacks []HeaderCallback
 	Client    *http.Client
+}
+
+type Resolution struct {
+	Data    []byte
+	Status  int
+	Headers http.Header
 }
 
 // Outbound client for http communication
@@ -43,7 +50,7 @@ func NewHttp(cfg Configuration) *Outbound {
 	}
 }
 
-func (h *Outbound) PostSingle(ctx context.Context, key string, payload []byte) (*http.Response, error) {
+func (h *Outbound) Single(ctx context.Context, key string, payload []byte) (*Resolution, error) {
 	path, err := url.JoinPath(h.baseURI, ofrepV1, key)
 	if err != nil {
 		return nil, fmt.Errorf("error building request path: %w", err)
@@ -59,5 +66,19 @@ func (h *Outbound) PostSingle(ctx context.Context, key string, payload []byte) (
 		req.Header.Set(callback())
 	}
 
-	return h.client.Do(req)
+	rsp, err := h.client.Do(req)
+	if err != nil {
+		return nil, err
+	}
+
+	b, err := io.ReadAll(rsp.Body)
+	if err != nil {
+		return nil, err
+	}
+
+	return &Resolution{
+		Data:    b,
+		Status:  rsp.StatusCode,
+		Headers: rsp.Header,
+	}, nil
 }
--- a/providers/ofrep/internal/outbound/http.go
+++ b/providers/ofrep/internal/outbound/http.go
@@ -17,27 +17,29 @@ const ofrepV1 = "/ofrep/v1/evaluate/flags/"
 type HeaderCallback func() (name string, value string)
 
 type Configuration struct {
-	Callbacks []HeaderCallback
 	BaseURI   string
+	Callbacks []HeaderCallback
+	Client    *http.Client
 }
 
 // Outbound client for http communication
 type Outbound struct {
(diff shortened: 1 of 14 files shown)


providers/ofrep/internal/outbound/http.go
```diff
@@ -0,0 +1,57 @@
...
+// Outbound client for http communication
+type Outbound struct {
+	auth    AuthCallback
+	baseURI string
+
+	client http.Client
+}
+
+func NewOutbound(baseUri string, callback AuthCallback) *Outbound {
+	client := http.Client{
```
liran2000, line 27: intermediate comment, though WIP:
Should expose the http client to the provider config ?
so consumer can have control of timeout, proxy etc'.

Kavindu-Dodan: Yeah, we can give some configurations here. I will add these with next commits

Kavindu-Dodan: @liran2000 I avoided adding any extra configurations in this first PR. I think we will need to add support for self-signed certificates and custom timeouts. But I think I will keep this as a follow-up PR

lukas-reining (approved):
Looks good to me! 
All seems as we planned it for JS and I can not find any issues.

providers/ofrep/internal/outbound/http.go
thisthat: I would also prefer having the possibility of passing through my own http client instead of providing a subset of its configuration. This will limit our maintenance effort to support all possible configurations.

Kavindu-Dodan: @thisthat this's a really good suggestion. I added `WithClient(*http.Client)` option with https://github.com/open-feature/go-sdk-contrib/pull/477/commits/a9e7df81a2a256015c504108e4343baa78ab62f8 allowing to override the http client with a custom one :)

Kavindu-Dodan: @liran2000 commit https://github.com/open-feature/go-sdk-contrib/pull/477/commits/a9e7df81a2a256015c504108e4343baa78ab62f8 fulfills your suggestion :)

thisthat (approved):

thomaspoignant:
Hey @Kavindu-Dodan great work, just one quick question.
I see a lot of similar code, and I was wondering why you did not use generics since you are using `go1.21`?
It will make the provider maybe a bit easier to understand.

What do you think?

thomaspoignant:
I also think that you don't have implement the `retry-later` header when we receive a `429`

Kavindu-Dodan:
@thomaspoignant thank you for reviewing the PR.

> I see a lot of similar code, and I was wondering why you did not use generics 

I assume you are referring to flags.go [1]. Yeah, it seems to have repeated logic differed by returned type, but each handler has value conversions and a specific return type. I had a quick look into refactoring them but I couldn't find any good way of doing that. 

> I also think that you don't have implement the retry-later header when we receive a 429

For Go SDK, I focused on the dynamic context paradigm, hence there is no support for bulk evaluation. So to handle 429, I think what we should do is to introduce a request blocker with a timer [2]. I haven't done it in this PR. 

[1] - https://github.com/open-feature/go-sdk-contrib/pull/477/files#diff-df3e1ced5c1dbe3274ff2d4cc8a47eaf0aab9475b4f526501c30491cb9d499ca
[2] - https://github.com/open-feature/go-sdk-contrib/pull/477/files#diff-4f6c2e021ab239118a7327dccc6fcab63a9f8480d79e34f57006ad7114fb463eR69-R70

thomaspoignant (approved):
👌

liran2000 (approved):

Allzz77123 (approved):

PR #528, opened 2024-06-14
feat(from_env): option for mapping flag to env name

### LINKED ISSUE #526: Env provider option for mapping flag names
Most shells doesn't allow env variables with dashes. Also, by convention, environment variables should be uppercased and words separated by underscores.
However, feature flags no always following this convention.
When you want to switch between multiple providers easily (and I believe this is one of the OpenFeature targets) your flags should have the same name in all providers, which is not the case for the env provider.

Perhaps this could be solved by providing a mapper function as a provider option. This mapper could map requested flag name into environment variable name.
Configuration solutions like viper and koanf already follow this approach for environment variable providers and it works fine

morigs:
## This PR

Provides a new option for mapping flag names to env variable names

### Related Issues

Resolves #526

### DIFF
--- a/providers/from-env/README.md
+++ b/providers/from-env/README.md
@@ -1,12 +1,12 @@
 # Environment Variable JSON Flag Provider
 
-This repository contains a very simple environment variable based feature flag provider.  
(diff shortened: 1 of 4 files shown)


Kavindu-Dodan (approved):
I think this is a good feature to help with specific usage requirements 👍

toddbaert (approved):
yepyepyep!

PR #530, opened 2024-06-14
feat(otel): add NewMetricsHook, use api interface instead of SDK type

### LINKED ISSUE #527: Otel metrics hooks constructor should accept MeterProvider from api, not SDK
`NewMetricsHookForProvider` expects `go.opentelemetry.io/otel/sdk/metric.MeterProvider` as a parameter. So as a user you have to have an access to the specific provider implementation, which is not always the case.

I suggest replacing the type of the first parameter to the `go.opentelemetry.io/otel/metric.MeterProvider` interface. It's not a breaking change because the current type implements this interface.

Also, we can provider a second constructor which will create a hook using otel global meter provider.

morigs:
## This PR

- Changes `NewMetricsHookForProvider` to accept MeterProvider interface from api instead of sdk type. This allows to pass any compatible implementation
- Adds `NewMetricsHook` which uses global default MeterProvider
- Tidies deps

### Related Issues

Fixes #527


toddbaert (approved):
Nice catch!

toddbaert:
I'll merge this in the next day or 2, but giving a bit of time for others to comment.

hooks/open-telemetry/pkg/metrics.go
```diff
@@ -2,10 +2,11 @@ package otel
 
 import (
 	"context"
+
 	"github.com/open-feature/go-sdk/openfeature"
+	"go.opentelemetry.io/otel"
```
Kavindu-Dodan, line 7: +1 for this change

Kavindu-Dodan (approved):

toddbaert:
Good enough for me.

PR #532, opened 2024-06-26
fix: Use LDClientInterface instead of *LDClient

zellyn:
It's often advantageous or necessary to wrap an `LDClient`, which
doesn't work if the provider code accepts a raw `*ld.LDClient`.

Signed-off-by: Zellyn Hunter <zellyn@gmail.com>

## This PR
Use `interfaces.LDClientInterface` instead of `*ld.LDClient` to refer to the wrapped LaunchDarkly client.

### Related Issues
n/a

### Notes
n/a

### Follow-up Tasks
n/a

### How to test
n/a — existing tests are sufficient


providers/launchdarkly/pkg/provider.go
```diff
@@ -47,11 +47,11 @@ func WithKindAttr(name string) Option {
 // Provider implements the FeatureProvider interface for LaunchDarkly.
 type Provider struct {
 	options
-	client *ld.LDClient
+	client interfaces.LDClientInterface
```
kinyoklion, line 50: I am skeptical of doing this. 

We are considering removing some of these interfaces in the future because it introduces breaking changes  that otherwise would not be breaking.

Because of the structural typing of Go you can locally make your own constrained interface for unit tests and things, and it won't break when we extend the API of the client. But if we extend an interface it does break such code.

zellyn: In our internal Go monorepo codebase, we have:

```
type metricsWrappedClient struct {
	*ld.LDClient
	counter statsd.TaggedCounter
}

var _ interfaces.LDClientInterface = (*metricsWrappedClient)(nil)

func wrapClientWithDefaultMetrics(client *ld.LDClient) *metricsWrappedClient {
...
```

We use that to emit custom metrics. Do the interfaces not exist for exactly that purpose?

kinyoklion: So that code would stop compiling if we added a function to LDClientInterface. As a result we have many methods not on the interface. So we add them to the client and next time we major version them we add them to the interface.

If you instead made an interface with only the functions you call, then there isn't any breaking.

kinyoklion: Correcting a comment. In your case wrapping here is fine. But the interface isn't only wrapping. If you implement the interface, then adding functions to the interface would break the code. 

If you just make an interface with the signature of the functions you use, then it can be implemented or wrapped and functions can be added and everything continues to work. 

We actually don't add methods to the LDClientInterface until we happen to make a major version, otherwise each addition is technically a break.

kinyoklion: As an aside to that, in our internal providers we have now moved to the provider taking a config and internally making a client. This allows it to work better with the provider initialization and shutdown which have been added to the spec. 

I think, with OpenFeature, if you want to wrap a client you should wrap the OpenFeature client and not the LaunchDarkly client. Though I see how it could be useful for the transitional code in an existing code base.

kinyoklion: Though you actually don't need to wrap either for this use case now. OpenFeature has hooks and you could use those.

LaunchDarkly also now has hooks and they could be used.

zellyn: Actually, on closer inspection, this should be `interfaces.LDClientEvaluations`, since that's all that this code uses.

> Because of the structural typing of Go you can locally make your own constrained interface for unit tests and things, and it won't break when we extend the API of the client. But if we extend an interface it does break such code.
Yes, this works great as long as you're confined to your own code. Trying to pass that constrained interface for unit tests through this provider and failing is exactly the kind of problem this PR is trying to fix. The general rule as I understand it is to return concrete structs (or pointers to them), but accept interfaces as parameters.

> We actually don't add methods to the LDClientInterface until we happen to make a major version, otherwise each addition is technically a break.
Absolutely. I'm pretty sure anyone implementing that interface (rather than a code-local subset as you described above) would be accustomed to that dance of having to add a new method when the upstream interface expands.


All that said: this code only actually uses five methods, which is even narrower than `LDClientInterface` or even `LDClientEvaluations`. Would you be more comfortable with the introduction of a private interface here with just those methods?
```
BoolVariationDetail(flagKey, ldCtx, defaultValue)
StringVariationDetail(flagKey, ldCtx, defaultValue)
Float64VariationDetail(flagKey, ldCtx, defaultValue)
IntVariationDetail(flagKey, ldCtx, int(defaultValue))
JSONVariationDetail(flagKey, ldCtx, ldvalue.CopyArbitraryValue(defaultValue))
```

kinyoklion: That sounds ideal to me.

zellyn:
@kinyoklion ptal — I defined a narrow `LDClient` interface.

Kavindu-Dodan (approved):

providers/launchdarkly/pkg/provider.go
Kavindu-Dodan: Looks good, added my approval :)

kinyoklion (approved):

Kavindu-Dodan:
Merging with current approvals :)

PR #538, opened 2024-07-26
fix: Should return a parse error if error while parsing

thomaspoignant:
## This PR
Should return a parse error if error while parsing


beeme1mr (approved):

Kavindu-Dodan (approved):
Good improvement 👍

PR #539, opened 2024-07-26
fix: Default GO parsing is float64 for int

thomaspoignant:
## This PR
By default GO when unmarshaling a JSON string GO is mapping all `int` as `float64`.
Before this PR we had only type_mismatch error for int.

### DIFF
--- a/providers/ofrep/internal/evaluate/flags.go
+++ b/providers/ofrep/internal/evaluate/flags.go
@@ -150,6 +150,18 @@ func (h Flags) ResolveInt(ctx context.Context, key string, defaultValue int64, e
 		value = int64(evalSuccess.Value.(int))
 	case int64:
 		value = evalSuccess.Value.(int64)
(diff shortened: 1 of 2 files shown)


beeme1mr (approved):

providers/ofrep/internal/evaluate/flags.go
```diff
@@ -150,6 +150,18 @@ func (h Flags) ResolveInt(ctx context.Context, key string, defaultValue int64, e
 		value = int64(evalSuccess.Value.(int))
 	case int64:
 		value = evalSuccess.Value.(int64)
+	case float64:
+		value = int64(evalSuccess.Value.(float64))
+		if float64(value) != evalSuccess.Value.(float64) {
```
Kavindu-Dodan, line 155: +1 for this check :)

Kavindu-Dodan (approved):

PR #542, opened 2024-07-29
feat(fo-feature-flag): add DisableCacheMetrics to disable metrics collector

mjudeikis:
…metrics

## This PR

- Adds `DisableCacheMetrics` to disable cache hit metric collection. Backchannel creates back traffic, which is not always needed or even creates additional workload. 

### Related Issues

Fixes #1234523

### DIFF
--- a/providers/go-feature-flag/pkg/provider.go
+++ b/providers/go-feature-flag/pkg/provider.go
@@ -5,6 +5,12 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"path"
+	"time"
+
 	"github.com/bluele/gcache"
 	"github.com/open-feature/go-sdk-contrib/providers/go-feature-flag/pkg/model"
 	of "github.com/open-feature/go-sdk/openfeature"
@@ -12,11 +18,6 @@ import (
 	"github.com/thomaspoignant/go-feature-flag/exporter"
 	"github.com/thomaspoignant/go-feature-flag/exporter/webhookexporter"
 	"github.com/thomaspoignant/go-feature-flag/ffcontext"
-	"io"
-	"net/http"
-	"net/url"
-	"path"
-	"time"
 )
 
 const defaultCacheSize = 10000
@@ -33,6 +34,7 @@ type Provider struct {
 	cache                  gcache.Cache
 	cacheTTL               time.Duration
 	cacheDisable           bool
+	dataCollectorDisable   bool
 	dataCollectorScheduler *exporter.Scheduler
 }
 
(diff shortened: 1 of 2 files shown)


providers/go-feature-flag/pkg/provider.go
```diff
@@ -328,6 +331,9 @@ func evaluateWithRelayProxy[T model.JsonType](provider *Provider, ctx context.Co
 			// call to convertCache wouldn't result in the same error on the next call.
 			provider.cache.Remove(cacheKey)
 		} else {
+			if provider.cacheMetricsDisable {
+				return cacheValue
+			}
```
thomaspoignant, line 336: The problem with this early return is that you are not changing the reason `cacheValue.Reason = of.CachedReason`. 

I think it may be better to put the if around the export.

Something like: 

```go
if !provider.cacheMetricsDisable {
 event := exporter.NewFeatureEvent(
				ffcontext.NewEvaluationContext(goffRequestBody.EvaluationContext.Key),
				flagName,
				cacheValue.Value,
				cacheValue.Variant,
				cacheValue.Reason == of.ErrorReason,
				"",
				"PROVIDER_CACHE",
 )
} ```

File: providers/go-feature-flag/pkg/provider.go [hunk 2]
```diff
@@ -328,6 +331,9 @@ func evaluateWithRelayProxy[T model.JsonType](provider *Provider, ctx context.Co
 			// call to convertCache wouldn't result in the same error on the next call.
 			provider.cache.Remove(cacheKey)
 		} else {
+			if provider.cacheMetricsDisable {
```
thomaspoignant, line 334: The problem with this early return is that you are not changing the reason `cacheValue.Reason = of.CachedReason`. 

I think it may be better to put the if around the export.

Something like: 

```go
if !provider.cacheMetricsDisable {
 event := exporter.NewFeatureEvent(
				ffcontext.NewEvaluationContext(goffRequestBody.EvaluationContext.Key),
				flagName,
				cacheValue.Value,
				cacheValue.Variant,
				cacheValue.Reason == of.ErrorReason,
				"",
				"PROVIDER_CACHE",
 )
provider.dataCollectorScheduler.AddEvent(event)
} 
```

providers/go-feature-flag/pkg/provider.go
```diff
@@ -33,6 +34,7 @@ type Provider struct {
 	cache                  gcache.Cache
 	cacheTTL               time.Duration
 	cacheDisable           bool
+	cacheMetricsDisable    bool
```
thomaspoignant, line 37: I would change the name to `dataCollectorDisable`, I did a wrong choice naming that cache in the beginning.

```suggestion
	dataCollectorDisable    bool
```

mjudeikis: Right :) if you change name it changes bit of other code too. Commited:
```
	if options.DisableCache || options.DisableDataCollector {
		return nil
	}
```

thomaspoignant (approved):

PR #546, opened 2024-07-30
feat(go-feature-flag): GO Feature Flag in process using GO module

thomaspoignant:
## This PR
In this PR we create a new provider for GO Feature Flag in process.

GO Feature Flag can be run as a go module and it was possible to run it like this in https://github.com/open-feature/go-sdk-contrib/tree/main/providers/go-feature-flag BUT this brings a lot of unused dependencies in the provider that are not used if you are using the relay-proxy.

In another PR we will release a refactor that removes the dependency to `GO Feature Flag` to minimize the provider's dependencies and keep it lightweight if it uses the relay proxy.

### Follow-up Tasks
Another PR will come with the refactor of https://github.com/open-feature/go-sdk-contrib/tree/main/providers/go-feature-flag.


thomaspoignant:
The test failing is in the `open-telemetry` hook, and from my early analysis, it may be because of the GO version (`1.22.5`).

toddbaert:
I'll take a look at this tomorrow and see what I can do abou t the CI issue.

toddbaert:
@thomaspoignant I fixed the CI issue with some dep updates (done in another branch and merged to main).

providers/go-feature-flag-in-process/go.mod
```diff
@@ -0,0 +1,22 @@
+module github.com/open-feature/go-sdk-contrib/providers/go-feature-flag-in-process
+
+go 1.22.5
```
toddbaert, line 3: of course it's up do you... but consider the go SDK still only requires 1.21, so you may be excluding some adoption with this.

providers/go-feature-flag-in-process/pkg/provider.go
```diff
@@ -0,0 +1,198 @@
...
+	if options.GOFeatureFlagConfig == nil {
+		return nil, fmt.Errorf("invalid provider options, empty GOFeatureFlagConfig value")
+	}
+
+	goff, err := ff.New(*options.GOFeatureFlagConfig)
+	if err != nil {
+		return nil, err
+	}
+	return &Provider{
+		goFeatureFlagInstance: goff,
```
toddbaert, line 34: So these are the same internals that power GoFF as a standalone server I guess? That's cool!

Do you plan to support in-process in other languages? I guess that would be a lot harder :sweat_smile:

toddbaert (approved):
:+1:

providers/go-feature-flag-in-process/go.mod
thomaspoignant: Yes I know, but the library is in 1.22.5 so I have no real choice here 😒

providers/go-feature-flag-in-process/pkg/provider.go
thomaspoignant: Yes exactly, this is the complete GO Feature Flag code behind the provider.

In-process for other languages is something I have in mind, but I haven't found the right approach for now, the rule library I use is mainly for GO only, and I am not sure how much effort it is to have the same compatibility level for other languages.

PR #548, opened 2024-08-07
ci: add z4kn4fein from ConfigCat as component owner

z4kn4fein:
## This PR
- Adds @z4kn4fein (ConfigCat) to the component owner's list of the ConfigCat provider. The intention behind this is that we (ConfigCat) would like to mark the provider in this repo as the official ConfigCat OpenFeature Go provider, and we would like to participate in its development and maintenance.

### DIFF
--- a/.github/component_owners.yml
+++ b/.github/component_owners.yml
@@ -7,6 +7,7 @@ components:
     - toddbaert
   providers/configcat:
     - rcrowe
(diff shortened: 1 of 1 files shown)


toddbaert (approved):

toddbaert:
Thanks @z4kn4fein .

PR #554, opened 2024-08-14
chore: add license to module

toddbaert:
Our license tooling recently started failing (due to updates) because our modules are missing licenses (we only have them at the root). This adds licenses to all modules.

### DIFF
--- a/hooks/open-telemetry/LICENSE
+++ b/hooks/open-telemetry/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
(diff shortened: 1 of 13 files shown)


kinyoklion (approved):

hooks/open-telemetry/LICENSE
```diff
@@ -0,0 +1,201 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
matthewelwell, line 189: Seems like more pain to maintain, but I suppose technically these should be updated with the open feature details, right? I see that this isn't the case in the root LICENSE file though so perhaps we can just ignore it.

thomaspoignant (approved):

z4kn4fein (approved):

PR #598, opened 2024-12-02
feat: Support supplying a custom sync provider for in-process flagd

cupofcat:
## This PR

- When the user is using in-process flagd, allows them to supply a `sync.ISync` implementation of their own that can fetch the flag config in any way they like.

### Notes

This PR provides the capability in Go that was recently added in [Java](https://github.com/open-feature/java-sdk-contrib/pull/900).

### DIFF
--- a/providers/flagd/pkg/configuration.go
+++ b/providers/flagd/pkg/configuration.go
@@ -2,10 +2,12 @@ package flagd
 
 import (
 	"fmt"
-	"github.com/go-logr/logr"
-	"github.com/open-feature/go-sdk-contrib/providers/flagd/internal/cache"
 	"os"
 	"strconv"
+
+	"github.com/go-logr/logr"
+	"github.com/open-feature/flagd/core/pkg/sync"
+	"github.com/open-feature/go-sdk-contrib/providers/flagd/internal/cache"
 )
 
 type ResolverType string
@@ -52,6 +54,8 @@ type providerConfiguration struct {
 	Selector                         string
 	SocketPath                       string
 	TLSEnabled                       bool
+	CustomSyncProvider               sync.ISync
+	CustomSyncProviderUri            string
 
 	log logr.Logger
 }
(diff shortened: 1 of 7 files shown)


providers/flagd/pkg/configuration.go
```diff
@@ -52,6 +54,8 @@ type providerConfiguration struct {
 	Selector                         string
 	SocketPath                       string
 	TLSEnabled                       bool
+	CustomSyncProvider               sync.ISync
+	CustomSyncProviderUri            string
```
alan-kut, line 58: Why do you need the URI as well?
Is it only to be returned from the makeSyncProvider?

It worries me a little bit because the URI here is not used at all. Caller needs to make sure they return a valid URI for the provider. At the same time most (all?) providers have the uri inside. Maybe it's possible to extend the interface to include a GetURI method?

cupofcat: URI is needed to set up the `FlagStore`. For other flavors of the in-process resolver, the URI is either the offline file path or is constructed by `makeSyncProvider` from host:port or explicitly provided `TargetUri`.

For a custom sync sync provider, because there is no default implementation, we could either try to do something clever (hard to do, given the implementation are all supposed to be custom) or just allow the user to explicitly setting it, which is what I went with.

I don't think changing the interface (which is defined in the core flagd repo) is necessary here. Also, wouldn't it potentially be a breaking change?

toddbaert: In the flagd daemon, the URI here is important because you can be source flags from multiple places... when you start the daemon, you can point it at `n` sources, like this: `./bin/flagd start --uri file:source-A.json --uri file:source-B.json --uri file:source-C.json`, and the URLs are used later as metadata when flags are evaluated, and in merging logic if flags are duplicated in sources, etc.

**However, in this case, I think it's irrelevant:** the in-process mode will ever only support one source (one iSync) and not more; so you could put anything in this URL in this case and it would work... it doesn't even need to be a proper URL; you could make it an optional string called "sourceName" or something like that, or even get it dynamically from the implementation using reflection. I think to make things less confusing, I would make this optional, or try to generate something for people, perhaps with `reflect.TypeOf`. They would only see it in some metadata nd logs, I think, unless I'm missing something.

I hope that makes sense, and helps a bit.

Otherwise I see no issues with this PR and I will happily accept it as soon as we agree on this.

beeme1mr (approved):

toddbaert (approved):

PR #607, opened 2025-01-04
fix: IntEvaluation when cannot be parsed

earlbread:
## This PR

- fixes `IntEvaluation` for Unleash provider.
- fixes `TestIntEvaluation` to run correctly and adds `int-flag` to `TestEvaluationMethods`.

### Related Issues

Related to https://github.com/open-feature/go-sdk-contrib/pull/465

PR #610, opened 2025-01-09
fix: IntEvalution, FloatEvalution when disabled or missing flags

earlbread:
## This PR

- fixes the `TYPE_MISMATCH` error in `IntEvaluation` and `FloatEvaluation` when flags are disabled or missing.
- fixes test cases for `IntEvaluation` and `FloatEvaluation`.
- fixes an issue where test cases are not reading the bootstrap file due to a backup file.

### Related Issues

Related to https://github.com/open-feature/go-sdk-contrib/pull/607

### DIFF
--- a/providers/unleash/pkg/provider_test.go
+++ b/providers/unleash/pkg/provider_test.go
@@ -40,27 +40,63 @@ func TestBooleanEvaluation(t *testing.T) {
 
 func TestIntEvaluation(t *testing.T) {
 	defaultValue := int64(0)
-	expectedValue := int64(123)
-	resolution := provider.IntEvaluation(context.Background(), "int-flag", defaultValue, nil)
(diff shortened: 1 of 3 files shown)


providers/unleash/pkg/provider_test.go
```diff
@@ -258,5 +298,6 @@ func TestMain(m *testing.M) {
 
 	cleanup()
 
+	os.Remove(backupFile)
 	os.Exit(exitCode)
```
liran2000, line 302: Maybe do it with defer?

earlbread: I tried defer, but it doesn't work because of `os.Exit(exitCode)'.

james-milligan: i think this makes sense - only other option is to put both in a single defer func but i dont see much value in this

liran2000 (approved):

beeme1mr (approved):

PR #626, opened 2025-02-11
feat(flagd): Support supplying providerID for in-process resolver as an option

cupofcat:
## This PR
- When the user is using in-process flagd, allows them to supply a providerID string to be used by grpc.Sync
- Adds tests and documentation for providerID and, already existing and tightly related but not fully documented/tested selector string


beeme1mr:
Could you please add `providerId` to the [provider spec](https://flagd.dev/reference/specifications/providers/#configuration) in flagd first?

cupofcat:
> Could you please add `providerId` to the [provider spec](https://flagd.dev/reference/specifications/providers/#configuration) in flagd first?

I can, but shouldn't we wait until this option is added to all the SDKs? How about we add this option to Go SDK only at first (this PR), and then add an issue to implement it in all the SDKs?

Re: [provider spec](https://flagd.dev/reference/specifications/providers/#configuration), I see two options:
1. hold off with updating the spec until all providers support it
2. add a column to the spec now that says which providers support the given option (would probably require a review of existing options as well)

Ideally, we should have tests that check for consistency and conformance to the spec across SKDs (unless that already exists?).

beeme1mr (approved):

beeme1mr:
I'll give folks a few more hours to review. If there are no objections, I'll merge and release a new update. Thanks @cupofcat.

By the way, I'll work on a PR to add `provideID` to the flagd spec. I believe it's just a miss since it's a useful property on the sync. We're trying to capture all the configurations options in the spec so they can be added to the [test suite](https://github.com/open-feature/flagd-testbed).

cupofcat:
@beeme1mr Thanks, sounds good!

Btw, I might be mistaken, but I think other SDKs currently don't allow for setting the providerID (neither from an env var, not from the constructor). Wouldn't adding this to test suite make the tests fail right now?

beeme1mr:
> I think other SDKs currently don't allow for setting the providerID (neither from an env var, not from the constructor). Wouldn't adding this to test suite make the tests fail right now?

You're right, it's missing from other Providers. That's why I'd like to add it to the spec and the test harness. The tests won't start failing until the test suite is updated for each Provider. This can help us detect inconsistencies and address them.

PR #644, opened 2025-03-05
feat!: updated OTel finally hook for upcoming changes

### LINKED ISSUE #642: Upgrade OTel hook
The hook interfaced in the Go SDK recently changed to align with the OpenFeature spec. This change was made [here](https://github.com/open-feature/go-sdk/pull/328) and will be available in the next release of the SDK. The OTel hook must be updated to support this change.

## Tasks

- Bump the version of the Go SDK in the OTel hook
- Update the [OTel metrics hook](https://github.com/open-feature/go-sdk-contrib/blob/4b04d0ee71a0b7faae3c156a655e1eed26381237/hooks/open-telemetry/pkg/metrics.go#L125)
- Update the tests

## Follow ups

- Explicitly mention the change in the release notes since it's breaking but allowed according to the spec.

bbland1:
## This PR

- bumped the `go-sdk` version to 1.14.1
- added `evalDetails` to the parameters passed in the `Finally` hook in `metrics` hook
- updated the `metrics_test` to have the new `evalDetails`

### Related Issues

Fixes #642 

### Notes

- not sure if there is a specific way to bump the `go-sdk` version to the unreleased version with the change to the `Hook` it currently throws an error where it is called in `metrics` because the type issue
```
var _ openfeature.Hook = &MetricsHook{}
```

- in the issue it states to explicitly mention the breaking but ok because of spec in the release notes, I am not sure if that is something I can do in this PR and I just can't find how in the contributing or if it is after the reviewing

### DIFF
--- a/hooks/open-telemetry/go.mod
+++ b/hooks/open-telemetry/go.mod
@@ -3,7 +3,7 @@ module github.com/open-feature/go-sdk-contrib/hooks/open-telemetry
 go 1.21
 
 require (
-	github.com/open-feature/go-sdk v1.11.0
+	github.com/open-feature/go-sdk v1.14.1
 	go.opentelemetry.io/otel v1.28.0
 	go.opentelemetry.io/otel/metric v1.28.0
 	go.opentelemetry.io/otel/sdk v1.28.0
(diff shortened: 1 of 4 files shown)


bbland1:
I see there is another issue about bumping the go-sdk to 1.15 over all because there has been some conflict. This PR still needs to bump to 1.15 which I'm happy to do now that it is out but curious if issue #686 needs to be finished so it is bumped for everyone for me to bump it up here and things to pass?

sahidvelji:
I don't think we need the `!` in the PR title, given that the version of this hook module is at 0.x.

beeme1mr (approved):

PR #648, opened 2025-03-12
feat(flagd)!: add file mode to flagd provider

raphael-wigoutschnigg-dt:
## This PR
- adds file resolver to flagd provider
- in-process resolver stays backwards compatible

### DIFF
--- a/providers/flagd/README.md
+++ b/providers/flagd/README.md
@@ -22,7 +22,8 @@ Flag evaluations take place remotely at the connected flagd instance.
 To use in this mode, set the provider to the `openfeature` global singleton as shown below (using default values which align with those of `flagd`)
 
 ```go
-openfeature.SetProvider(flagd.NewProvider())
+provider, err := flagd.NewProvider()
+openfeature.SetProvider(provider)
 ```  
 
 ### In-process resolver
@@ -33,10 +34,11 @@ Flag configurations for evaluation are obtained via gRPC protocol using [sync pr
 Consider following example to create a `FlagdProvider` with in-process evaluations,
 
 ```go
-provider := flagd.NewProvider(
+provider, err := flagd.NewProvider(
         flagd.WithInProcessResolver(),
         flagd.WithHost("localhost"),
-        flagd.WithPort(8013))
+        flagd.WithPort(8013),
+)
 openfeature.SetProvider(provider)
 ```
 
@@ -50,9 +52,10 @@ The custom sync provider must implement the [sync.ISync interface](https://githu
 ```go
 var syncProvider sync.ISync = MyAwesomeSyncProvider{}
 
-provider := flagd.NewProvider(
+provider, err := flagd.NewProvider(
         flagd.WithInProcessResolver(),
-        flagd.WithCustomSyncProvider(syncProvider))
+        flagd.WithCustomSyncProvider(syncProvider),
+)
 openfeature.SetProvider(provider)
 ```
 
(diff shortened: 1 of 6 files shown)


providers/flagd/README.md
```diff
@@ -33,30 +34,16 @@ Flag configurations for evaluation are obtained via gRPC protocol using [sync pr
 Consider following example to create a `FlagdProvider` with in-process evaluations,
 
 ```go
-provider := flagd.NewProvider(
+provider, err := flagd.NewProvider(
         flagd.WithInProcessResolver(),
         flagd.WithHost("localhost"),
-        flagd.WithPort(8013))
+        flagd.WithPort(8013),
```
aepfli, line 40: hmm inprocess uses default port 8015 should we use this port here?

aepfli (approved):
Thank you this changes, especially the normalization in the readme look really good. But as we change the API, we need to add a `!` after `feat(flagd)` in the pr title to signal a breaking change.

providers/flagd/README.md
aepfli: i would stick with the defaults i think.... i case of doubt, we always check what the java provider is doing :D

aepfli (approved):

providers/flagd/pkg/provider.go
```diff
@@ -90,11 +91,39 @@ func NewProvider(opts ...ProviderOption) *Provider {
...
+		})
 	}
 
 	provider.service = service
 
-	return provider
+	return provider, nil
+}
+
+func UpdateProvider(p *Provider) {
```
toddbaert, line 105: I think this name could be better (and should almost certainly be lower-cased; in Go, upper case means a token is exported, and lower case means it's private).

Consider: `configureResolver` or something like that.

providers/flagd/pkg/provider.go
```diff
@@ -90,11 +91,39 @@ func NewProvider(opts ...ProviderOption) *Provider {
...
+		switch p.providerConfiguration.Resolver {
+		case rpc:
+			p.providerConfiguration.Port = defaultRpcPort
+		case inProcess:
+			p.providerConfiguration.Port = defaultInProcessPort
+		}
+	}
+}
+
+func CheckProvider(p *Provider) error {
```
toddbaert, line 120: Similar to https://github.com/open-feature/go-sdk-contrib/pull/648/files#r1999404241, consider `validate` or something like that.

providers/flagd/README.md
```diff
@@ -65,28 +48,44 @@ The custom sync provider must implement the [sync.ISync interface](https://githu
...
 
 > [!IMPORTANT]
 > Note that the in-process resolver can only use a single flag source.
 > If multiple sources are configured then only one would be selected based on the following order of preference:
 >   1. Custom sync provider
->   2. Offline file
->   3. gRPC
+>   2. gRPC
+
+### File mode
```
toddbaert, line 75: Thanks for doc updates :smile:

toddbaert (changes requested):
My only blocking concern is that I think we are exporting some functions that seem like they should be private (see comments).

If I'm mistaken, please let me know.

providers/flagd/pkg/provider.go
raphael-wigoutschnigg-dt: You are right. Exporting was not needed - I was under the wrong impression this is needed for testing purposes.

toddbaert (approved):

PR #651, opened 2025-03-19
test(flagd): adding configuration gherkin tests

raphael-wigoutschnigg-dt:
## This PR
adding configuration gherkin tests


providers/flagd/pkg/provider.go
```diff
@@ -235,147 +195,147 @@ func (p *Provider) setStatus(status of.State) {
 
 // ProviderOptions
 
-type ProviderOption func(*Provider)
+type ProviderOption func(*ProviderConfiguration)
```
aepfli, line 198: should those methods also be part of the configuration.go? as they are really configuration related?

tests/flagd/pkg/integration/config.go
```diff
@@ -0,0 +1,293 @@
...
+	},
+	"offlineFlagSourcePath": {
+		currentValue: func(config *flagd.ProviderConfiguration) interface{} {
+			return config.OfflineFlagSourcePath
+		},
+	},
+}
+
+// providerOptionGeneratorMap a map that defines the provider options generators for supported options
+var providerOptionGeneratorMap = map[string]func(value string) (flagd.ProviderOption, error){
```
aepfli, line 100: do we really need this mapping? we can access the properties of the providerConfiguration directly - the helper methods are nice and simulate a more go like way, but i think we can run this test with directly setting the properties, wdyt? we even could set those fields via reflection

aepfli (commented):
thank you, this is the foundation for our gherkin tests and it is really really cool. I am not sure if we should opt in for reflection and not utilize the helper methods, what do you think?

aepfli (approved):

toddbaert:
I gave a check over, and though we changed some names of internal struct fields, we didn't make any breaking changes here to config methods or publicly exposed members (I think) . That's my only minor worry about this PR. Please confirm and I will approve.

raphael-wigoutschnigg-dt:
@toddbaert I checked the change again. `ProviderConfiguration` is now exported (to be able to test the config) and was changed slightly (`cache`, `tls`). Because it was internal prior to this change I don't see an issue wrt breaking changes. In addition `NewProviderConfiguration` was added - which is not needed by consumers of the sdk (feels like a design flaw - from testing perspective).

I would say: No breaking change but - from production perspective - unnecessary types/func exposed.

toddbaert:
> @toddbaert I checked the change again. `ProviderConfiguration` is now exported (to be able to test the config) and was changed slightly (`cache`, `tls`). Because it was internal prior to this change I don't see an issue wrt breaking changes. In addition `NewProviderConfiguration` was added - which is not needed by consumers of the sdk (feels like a design flaw - from testing perspective).
> 
> I would say: No breaking change but - from production perspective - unnecessary types/func exposed.

I think it's worth it in this case for independent testing of the config.

@aepfli what do you think?

aepfli:
I think exporting this endpoint allows further test-harness steps. as we cannot utilize just a configuration object and pre-create it for our tests.

PR #716, opened 2025-06-29
fix: fix goroutine leaks around shutdown

bduffany:
Ensure that Shutdown immediately terminates all goroutines and waits for them to exit.

This fixes a goroutine leak in the case where the `events` channel buffer is full, and no goroutines are actively receiving from it.

Even in the case where the `events` buffer is not full though, `Shutdown()` should synchronously wait for all goroutines to exit, rather than shutting them down and then letting them exit asynchrously. Async shutdown is not desirable, especially for tests, because it means that if (e.g.) some log messages are happening during shutdown, then those log messages will be potentially attributed to the wrong test case, which is confusing.

There are two fixes:

1. Add a `sync.WaitGroup` to track any started goroutines, and call `Wait()` on it during shutdown.
2. Ensure that ctx cancellation is always respected when doing any blocking operations, to ensure that shutdown is both immediate and guaranteed. Particularly, respect cancellation for both (A) sleeping (during exponential backoff) and (B) all channel operations. This not only fixes a goroutine leak when `c.events` is full, but also ensures that shutdown happens immediately when it is requested.

https://github.com/open-feature/go-sdk/issues/397 is a related issue, in the sense that the goal of this change is to improve testability and ensure no goroutine leaks.


providers/flagd/pkg/service/in_process/service.go
```diff
@@ -93,15 +94,19 @@ func (i *InProcess) Init() error {
 	syncChan := make(chan sync.DataSync, 1)
 
 	// start data sync
+	i.wg.Add(1)
 	go func() {
+		defer i.wg.Done()
```
chrfwow, line 99: I am not too familiar with go. Could it be possible that the goroutine is not started after `i.wg.Add(1)` because of shutdown being called, and therefore `i.wg.Done()` is never executed?

beeme1mr (approved):

providers/flagd/pkg/service/in_process/service.go
chrfwow: Thanks for the detailed explanation!

chrfwow (approved):

toddbaert:
Thanks a lot @bduffany !

PR #726, opened 2025-07-19
feat(aws-ssm): implement provider

gdegiorgio:
This PR

 Implements a new server provider for AWS System Manager, matching the same implementation already present in [JS SDK](https://github.com/gdegiorgio/js-sdk-contrib/tree/main/libs/providers/aws-ssm) (See https://github.com/open-feature/js-sdk-contrib/pull/1221)


sahidvelji (approved):
Thanks for your contribution, and your patience!

PR #741, opened 2025-09-06
test(flagd): add envoy tests for targetUri

aepfli:


aepfli:
interesting finding rpc does not support targeturi ;)

PR #744, opened 2025-09-07
feat(flagd): add eventing with graceperiod fir inprocess resolver

aepfli:
this is a follow up to #743


aepfli:
@open-feature/flagd-approvers please take a look at this, the internals have changed a little - but from the outside this should not be breaking. We are not conform with the provider spec for inprocess when it comes to events and error handling, imho a big improvement, also we rely now on grpc for the retry handling. please let me know what you think about this

tangenti:
This is a large change. Could you add more details in the PR description for people to understand the problem it's trying to solve here?

aepfli:
> Thanks Simon. For events and grace periods, are those documented somewhere? Is it already implemented in the flagd providers of other languages?
> 
> CC @andreyturkov

it is documented [here](https://flagd.dev/reference/specifications/providers/#flagd-provider-lifecycle) and [here](https://github.com/open-feature/flagd-testbed/blob/main/gherkin/events.feature) we do have gherkin tests for our e2e tests, and they are executed as long as the `@grace` filter is not disabled. we do have those for java and python for sure - i would need to check for the others - but it is a defined behaviour.

PR #746, opened 2025-09-15
ci: upgrade golangci-lint to v2

erka:
## This PR

- upgrade golangci-lint to v2
- address minor issues reported by updated linters

### Notes

The motivation is to keep the same version and configuration of golangci-lint between the `go-sdk` and `go-sdk-contrib` repositories.

Some linters and rules have been temporarily disabled to simplify the upgrade process and will be re-enabled separately.

### DIFF
--- a/providers/aws-ssm/provider.go
+++ b/providers/aws-ssm/provider.go
@@ -11,6 +11,8 @@ type Provider struct {
 	svc *awsService
 }
 
+var _ openfeature.FeatureProvider = (*Provider)(nil)
+
 type ProviderOption func(*Provider)
 
 func NewProvider(cfg aws.Config, opts ...ProviderOption) (*Provider, error) {


beeme1mr (approved):

providers/multi-provider/pkg/providers.go
```diff
@@ -36,9 +36,9 @@ type (
 		customStrategy   strategies.Strategy
 		logger           *slog.Logger
 		publishEvents    bool
-		metadata         *of.Metadata
+		metadata         *of.Metadata //nolint unused
```
thomaspoignant: if it is unused should we keep the configuration?

providers/prefab/pkg/provider_config.go
```diff
@@ -8,5 +8,5 @@ type ProviderConfig struct {
 	// EnvironmentNames             []string
 	// ProjectEnvID                 int64
 	// InitializationTimeoutSeconds float64
-	// OnInitializationFailure      OnInitializationFailure
+	//nolint OnInitializationFailure      OnInitializationFailure
```
thomaspoignant: Should we keep this if it is unused ?

providers/prefab/pkg/provider_test.go
```diff
@@ -363,7 +365,12 @@ func TestMain(m *testing.M) {
 		os.Exit(1)
 	}
 
-	of.SetProviderAndWait(provider)
+	err = of.SetProviderAndWait(provider)
+	if err != nil {
+		fmt.Printf("Error during provider set: %v\n", err)
+		os.Exit(1)
```
thomaspoignant: Do we really want to crash the test with `os.Exit` ?
It will probably better to return assert here rather than crashing the full test suite.

erka: This is `TestMain(m *testing.M)` and from [docs](https://pkg.go.dev/testing#hdr-Main)

>> `m.Run` will return an exit code that may be passed to [os.Exit](https://pkg.go.dev/os#Exit)

providers/unleash/pkg/provider_test.go
```diff
@@ -254,18 +258,19 @@ func cleanup() {
 	provider.Shutdown()
 }
 
-func TestMain(m *testing.M) {
+//go:embed demo_app_toggles.json
+var appToggles []byte
```
toddbaert: I'm curious if the linter recommended this or if you just found it.

PR #752, opened 2025-10-18
chore(flipt): bump minimum Go requirement to 1.24 and modernize

erka:
## This PR

- Upgrade Go version from 1.23.0 to 1.24.0
- Replace context.Background() with t.Context() in tests
- Use 'any' instead of 'interface{}' for cleaner syntax


sahidvelji (approved):
Thanks! Going to merge this as these are effectively just style changes.

PR #758, opened 2025-10-21
feat(launchdarkly): Implement openfeature.StateHandler interface for clean shutdowns

dangra:
## This PR

Implements [openfeature.StateHandler](https://pkg.go.dev/github.com/open-feature/go-sdk@v1.16.0/openfeature#StateHandler) interface to call [LDClient.Close](https://pkg.go.dev/github.com/launchdarkly/go-server-sdk/v7#LDClient.Close) on shutdown.

### How to test

It ensures the provider implements the `openfeature.StateHandler` interface the same way it is done for `openfeature.FeatureProvider` interface.

### DIFF
--- a/providers/launchdarkly/pkg/provider_test.go
+++ b/providers/launchdarkly/pkg/provider_test.go
@@ -302,7 +302,7 @@ func TestShutdown(t *testing.T) {
 		assert.Ok(t, err)
 
 		openfeature.Shutdown()
-		assert.Cond(t, !mockClient.closeCalled, "expected client.Close() to be called")
+		assert.Cond(t, !mockClient.closeCalled, "expected client.Close() not to be called")
 	})
 
 	t.Run("should call client close on shutdown", func(t *testing.T) {


erka:
Hey @dangra. Thank you for your contribution.

Could you share a bit more about why you want to do this? It seems a bit unfair that `StateHandler` is used this way. Since the provider gets the client from outside, I’d expect the client to be closed externally as well.

dangra:
hi @erka, that is a fair observation. 

My take is that I don't always setup LaunchDarkly as provider but I want to my code to be independent of the provider and be able to call `openfeature.Shutdown()` when stopping the service.

high level it looks like this:

```go
func main() {
   setupOpenfeature(...)
   defer openfeature.Shutdown()
  
  ofc := openfeature.GetApiInstance().GetClient()
  if ofc.Boolean("flag-name") {
      callSomeStuff()
  }
}

func setupOpenfeature(...) {
  switch envvarOrConfig {
  case "launchdarkly":
    ldc, _ := ldclient.MakeClient(sdkey, 0)
    provider := launchdarkly.NewProvider(ldc)
    openfeature.SetProvider(provider)
  case "other":
     // setup some other for local development or tests
  }
}
```

That is a simplified structure, my setup function does a lot more open-feature related. 

In fact, due to the recent extended outage caused by AWS that also hit LaunchDarkly, I have setup a LaunchDarkly Relay Proxy and with the use of [Multi-Provider](https://github.com/open-feature/go-sdk-contrib/tree/main/providers/multi-provider) it points to upstream and the relay.  I am mentioning this because it is two LD providers wrapped in a single provider that supports the [StateHandler](https://github.com/open-feature/go-sdk-contrib/blob/81cf5913281dd63e58b8e33bf526e761c8b53a6e/providers/multi-provider/pkg/providers.go#L263-L276) interface and works transparently.

`[openfeature.Shutdown](https://pkg.go.dev/github.com/open-feature/go-sdk/openfeature#Shutdown)` and `[LDClient.Close](https://pkg.go.dev/gopkg.in/launchdarkly/go-server-sdk.v5#LDClient.Close)` are both supposed to be called as last thing before shutting down, they are pretty much aligned by definition.

For the sake of completeness, this is my workaround in case other needs it:

```go
// The launchdarkly openfeature provider doesn't implement StateHandler interface
type LaunchdarklyProvider struct {
	*ofld.Provider
	ldClient *ld.LDClient
}

func (p *LaunchdarklyProvider) Init(evCtx openfeature.EvaluationContext) error {
	return nil
}

func (p *LaunchdarklyProvider) Shutdown() {
	_ = p.ldClient.Close()
}

func newLaunchdarklyProvider(ldClient *ld.LDClient) *LaunchdarklyProvider {
	return &LaunchdarklyProvider{
		ldClient: ldClient,
		Provider: ofld.NewProvider(ldClient),
	}
}
```

erka:
> @erka I won't argue against your approach, I think I understand the motivation behind it. To me in practical terms it is about passing the `ldclient.Client` instance ownership to the openfeature provider and forget about it. I also understand this change is backwards incompatible and may surprise current and future users.
> 

@dangra Yeah... someone could be unhappy

> The only alternative I see other than closing this PR and if you are up to, is to have a `NewProvider(...)` option named `WithCloseOnShutdown()` (or something else) that enables the new behavior so users can opt-in.
> 
> would that be ok with you?

Yes, this could be a solid middle-ground solution.

sahidvelji (approved):

PR #802, opened 2025-11-20
feat(optimizely): implement optimizely provider

JonCanning:
## This PR
- Adds a provider for [Optimizely Feature Experimentation](https://docs.developers.optimizely.com/feature-experimentation/docs/introduction)

### Variable resolution rules

- **No variables**: only `BooleanEvaluation` is supported, returns `Enabled` state. All other evaluation methods return an error.
- **Single variable**: returns the value of the only variable if the type matches. Type mismatch returns an error.
- **Multiple variables**: only `ObjectEvaluation` is supported, returns a map of variable names to values. All other evaluation methods return an error.

### Shutdown behaviour

The provider does not close the Optimizely client by default on shutdown, since the client may be shared. Use `WithCloseOnShutdown()` to opt in to closing the client when the provider shuts down.


JonCanning:
Thanks for the feedback and apologies for the quality issues, I was a bit too "Jesus take the wheel" with AI. I have reworked it so it's all my fault now 🙂

sahidvelji:
Thanks for this contribution! And, thank you @erka for the excellent feedback.

I noticed that there's an Optimizely Provider implementation in java-sdk-contrib, where the authors of that Provider have taken a different approach. For example, the Java Provider does not even support string, int, or float evaluations. I am not convinced that that's the right strategy though. Mapping vendor SDK concepts to OpenFeature concepts is not always straightforward. There are often multiple ways of doing so, each with their own trade-offs. Before accepting this contribution, I first really want to come to a consensus on the design of this Provider. I will try to get some more eyes on this.

https://github.com/open-feature/java-sdk-contrib/tree/main/providers/optimizely

sahidvelji:
Sorry for the delay. This Provider is of particular interest to me because my org has built a custom internal Provider that wraps Optimizely. We spent quite a bit of time deliberating various design decisions. It's tough because Optimizely's flags are quite different from other vendors in the sense that a single flag key does not always map to a single value.

Here are my thoughts.

Optimizely consists of projects, where a project contains flags. Flag keys must be unique across a project, and flags in Optimizely can have zero or more variables. There are three main cases to consider when mapping these to OpenFeature: a flag with zero variables, a flag with one variable, and a flag with more than one variable. The OpenFeature evaluation methods only accept a flag key and return a single value. To solve for that, we considered several approaches:

1. Use dot-delimited structured access for flag keys. Something like `flag_key.variable_key`. This felt like a hack to us, and didn't seem to be aligned with how OpenFeature generally works. For example, this wouldn't work well with other OpenFeature tooling like the CLI.
2. Use the evaluation context to store the variable key, as is done in this PR. We decided against this approach also because if we ever migrate to a different feature flag platform, their Provider wouldn't have that functionality and client code would need to be modified. This sort of defeats the purpose of using OpenFeature, given that one of the benefits is a vendor-neutral SDK. Also, I'm not sure that this is the intended purpose of the evaluation context.
3. Wrap the OpenFeature SDK by implementing the same interface, but also including an extra method to handle evaluation of flags with multiple variables. For the same reason as above, this approach did not make sense to us since introducing a wrapper defeats the purpose of using OpenFeature.
4. Instead of wrapping the OpenFeature SDK, add a method to the Provider that handles evaluation of flags with multiple variables. Also not preferred since this isn't how OpenFeature is intended to be used.

Here's what we landed on:

For flags with no variables, we support only the bool evaluation method and the value of `isFeatureFlagEnabled` is returned.
All other evaluation methods return an error.

For flags with a single variable, we support the evaluation method based on the type and return the value of the only variable in the flag. If the type doesn't match, return an error.

For flags with multiple variables, we use the ObjectEvaluation method, which returns an object of variables names to their values. All other evaluation methods return an error.

There is a problematic case where a flag has a single variable and a developer attempts to add another variable. As an example, this would cause a string evaluation to suddenly error because it now has multiple variables. Since this case is not possible as long as the flag is "running" in production, we decided this is OK.

Let me know where I can clarify our approach.

JonCanning:
Thanks @sahidvelji, that's very helpful, I'll update it with your suggestions

providers/optimizely/README.md
```diff
@@ -0,0 +1,114 @@
...
+    }).Client()
+    if err != nil {
+        panic(err)
+    }
+
+    provider := optimizely.NewProvider(optimizelyClient)
+    openfeature.SetProviderAndWait(provider)
+    defer openfeature.Shutdown()
+
+    ofClient := openfeature.NewClient("my-app")
```
sahidvelji:
```suggestion
    ofClient := openfeature.NewDefaultClient()
```

toddbaert:
Hey @JonCanning ! I'll give this a review tomorrow and merge it unless @sahidvelji or @erka have objections.

toddbaert (approved):
Overall the approach looks good. I agree with @erka 's comment about [closing](https://github.com/open-feature/go-sdk-contrib/pull/802#issuecomment-3671482082).

Please note, unless you want this to be releases as a `1.0` (I wouldn't recommend it) you need to add and entry here as well: https://github.com/open-feature/go-sdk-contrib/blob/main/.release-please-manifest.json.

I"m approving because we're on the right track, but I will wait for another from @erka or @sahidvelji before I merge.

erka (approved):

CraigAstillRVU:
We would be interested in using this provider. Is there a blocker to it being merged?

erka:
@CraigAstillRVU it has been released https://pkg.go.dev/github.com/open-feature/go-sdk-contrib/providers/optimizely

@JonCanning thank you for your work!

PR #815, opened 2026-01-06
fix(flagd): update flagd/core and remove vulnerable dependencies [security]

kuznero:
## Summary
- Update `github.com/open-feature/flagd/core` to v0.13.1/v0.13.2 to address security vulnerabilities
- Remove unused transitive dependencies with known vulnerabilities:
  - `github.com/xeipuuv/gojsonpointer`
  - `github.com/xeipuuv/gojsonreference`
  - `github.com/xeipuuv/gojsonschema`
  - `github.com/zeebo/xxh3`
  - `github.com/klauspost/cpuid/v2`
- Adapt InProcess service to handle the new `SetState` signature that no longer returns change information
- Flag changes are now computed by comparing store state before and after updates

Fixes #814

## Additional Changes
- Fix GOBIN path resolution in Makefile to handle multiple GOPATH entries
- Move `golang.org/x/exp` from direct to indirect dependency in flagd provider
- Improve test cleanup with `t.Cleanup` and graceful server shutdown
- Update flagd-http-connector to use flagd/core v0.13.2 and Go 1.24.0

## Test plan
- [x] Existing tests pass
- [x] Manual verification of in-process provider functionality

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ ALL_GO_MOD_DIRS := $(shell find . -type f -name 'go.mod' -exec dirname {} \; | s
 MODULE_TYPE ?= providers
 FLAGD_TESTBED = flagd-testbed
 FLAGD_SYNC = sync-testbed
-GOLANGCI_LINT_VERSION := v2.7.2
+GOLANGCI_LINT_VERSION := v2.8.0
 GOBIN := $(or $(shell go env GOBIN),$(shell go env GOPATH | cut -d: -f1)/bin)
 
 workspace-init:


kuznero:
The PR is ready for review by maintainers. Please approve the CI workflow run.

erka:
It would be helpful to understand why this was bumped in open-feature/flagd#1840, and whether it could be lowered back to 1.24.0 as it was before. cc @toddbaert

toddbaert:
Ya - flagd is a standalone app, so I wasn't very conservative with that update, forgetting that we consume it here as well...

That said, Go 1.26 is supposed to come out next month, which will make 1.24 EOL by the Go policies. Should we just wait top merge this fix until after then?

WDYT? @kuznero @erka ? If there's 1.24.x fixes for all the CVEs in https://github.com/open-feature/flagd/pull/1840, I'm fine with using that (I'm not 100% sure there is) for now, if it's really needed.

erka:
>  I'm fine with using that for now, if it's really needed (but I'm not 100% sure all those CVEs are addressed in 1.24.x, I will have to double-check).

Different teams have their own security policies and must handle vulnerabilities according to defined timelines. They may not be able to wait. This also would require people to upgrade to at the latest go (1.25.5), which may not be ideal when applying security fixes. @toddbaert

connyay:
I closed mine. Sorry.

PR #830, opened 2026-02-04
Add CustomHeaders option

olvlvl:
## This PR
Adds support for custom headers for every HTTP request, similar to the [option found on the Ruby SDK](https://github.com/open-feature/ruby-sdk-contrib/blob/413cfd8f65e62dc14bd514001e32f1b140dcdf3e/providers/openfeature-go-feature-flag-provider/lib/openfeature/go-feature-flag/options.rb#L16). Custom headers can be used to define a `User-Agent`, to prevent requests from being blocked by servers that require browser-like identification or specific device signatures.

### Notes
Our WAF requires the `User-Agent` header to be defined. Requests to the relay-proxy are being denied because the header is not specified.


olvlvl:
I can't edit the title, I'm opening another PR.

PR #868, opened 2026-04-16
feat(go-feature-flag): add caching support for remote evaluation

thomaspoignant:
## Description

- **What was the problem?** The remote evaluation mode (OFREP) had no caching layer, meaning every flag evaluation triggered a network call to the GO Feature Flag relay proxy.
- **How is it resolved?** An LRU cache (backed by `gcache`) is added to the `Remote` evaluator. Flags that include `gofeatureflag_cacheable: true` in their response metadata are stored in the cache. A background polling goroutine checks the relay proxy's `/v1/flag/configuration` endpoint using ETags and purges the cache automatically when the configuration changes.
- **How can we test?** New integration-style tests in `provider_test.go` cover cache hits/misses, LRU eviction, TTL expiration, non-cacheable flags, and disabled-cache mode. Hook-level tests verify that only cached evaluations are forwarded to the data collector (with source `PROVIDER_CACHE`).
- **Breaking changes:** None — caching is opt-in per flag via the `gofeatureflag_cacheable` metadata key, and can be disabled globally with `DisableCache: true`.

## Checklist

- [ ] I have read the **CONTRIBUTING** document.


hairyhenderson:
Is there a reason to call a GOFF-specific API rather than OFREP? IIRC the old implementation only used OFREP and didn't poll but rather cached for 1min regardless.

In our case the GOFF API isn't always reachable (while the OFREP API is), so this new mode won't work.

Any reason not to go back to the functionality that existed in https://github.com/open-feature/go-sdk-contrib/pull/851?

Though, I suppose perhaps it may make even more sense to just add that caching to the `ofrep` provider instead?

hairyhenderson:
after chatting on Slack, looks like I'd [misunderstood](https://gophers.slack.com/archives/C029TH8KDFG/p1776455038776839?thread_ts=1774979790.056459&cid=C029TH8KDFG)!

So, this LGTM!

PR #878, opened 2026-05-03
feat(providers/vercel): Add Vercel Flags provider

LukeSheard:
## This PR
Implements a Vercel Flags provider for the [Flags SDK](https://flags-sdk.dev/). It's based on the open-feature adapter in the [flags-sdk repository](https://github.com/vercel/flags/blob/main/packages/vercel-flags-core/src/openfeature.make.ts) which is in JS. 

### Notes
I'm still testing this myself, but opening a PR to gather feedback & raise my intentions about trying to support this.

### DIFF
--- a/providers/vercel/LICENSE
+++ b/providers/vercel/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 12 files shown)


erka:
Hey @LukeSheard 

Thank you for your PR.

Please don't use `pkg` folder/package. Put `provider.go`, `options.go` in the parent folder next to the `go.mod` and hidden non-public things into `internal` package like `xxhash.go`, `evlauator.go`, etc...

```go
import (
  "github.com/open-feature/go-sdk-contrib/providers/vercel"
)

func main() {
	provider, err := vercel.NewProvider()
}
```

Issue #789, opened 2025-11-05
[flagd] Update selector handling for consistency with sync, evaluation, and OFREP services


aepfli:
### Context
This ticket follows open-feature/flagd#1814, which aims to normalize selector usage across all flagd services. Currently, selectors are handled differently in sync, evaluation, and OFREP services.

### Requirements
- Reference: https://github.com/open-feature/flagd/issues/1814
- Ensure selector handling is consistent between go-sdk-contrib and core flagd, especially for clients and plugins interfacing with flagd.
- Ensure backwards compatibility, supporting the deprecated selector until the deprecation period ends.

### Tasks
1. Audit all usages of selector in go-sdk-contrib (including SDK components interacting with flagd)
2. Update to prefer `flagd-selector` header when communicating with flagd services
3. Ensure backward compatibility where needed
4. Document selector handling patterns for maintainers and users
5. Prepare for future removal of deprecated behaviors

### Benefits
- Predictable and consistent selector support when integrating with flagd
- Lower maintenance burden for migration in future major releases

### Additional notes
- Work in alignment with the migration guidance and deprecation plan discussed in the core flagd ticket.

PR #895, opened 2026-06-01
feat: make use of flagd-selector header in RPC mode

toddbaert:
- The in-process resolver already sends the `flagd-selector` gRPC metadata header, but the RPC resolver did not. This PR wires it up: the RPC resolver now sends the header on every call when a selector is configured, so the flagd server can serve the requested flag set.
- Brings the Go RPC resolver to parity with the Java, JS, .NET, and Python providers, which all already wire this header up (see flagd issue [#1814](https://github.com/open-feature/flagd/issues/1814)).
- Additive: behavior is unchanged when no selector is configured.

We missed the RPC mode in: #789

### DIFF
--- a/providers/flagd/internal/header/header.go
+++ b/providers/flagd/internal/header/header.go
@@ -1,0 +1,5 @@
+// Package header defines metadata header keys shared by the flagd resolvers.
+package header
+
+// Selector is the metadata header used to pass the flag-set selector to flagd.
+const Selector = "flagd-selector"
--- a/providers/flagd/pkg/service/rpc/interceptors.go
+++ b/providers/flagd/pkg/service/rpc/interceptors.go
@@ -4,7 +4,7 @@ import (
 	"context"
 
 	"connectrpc.com/connect"
-	"github.com/open-feature/go-sdk-contrib/providers/flagd/internal/header"
+	"github.com/open-feature/go-sdk-contrib/providers/flagd/internal/flagdmeta"
 )
 
 // selectorInterceptor is a connect.Interceptor that adds the flagd-selector header.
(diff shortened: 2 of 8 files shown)


providers/flagd/pkg/service/rpc/interceptors.go
```diff
@@ -0,0 +1,44 @@
...
+	"context"
+
+	"connectrpc.com/connect"
+)
+
+const SelectorMetadataKey = "flagd-selector"
+
+// selectorInterceptor is a connect.Interceptor that adds the flagd-selector
+// header to outgoing unary and streaming client requests when a selector is
+// configured. This mirrors the behavior of the in-process gRPC interceptor.
```
toddbaert:
```suggestion
// connect.Interceptor that adds the flagd-selector
```

providers/flagd/pkg/service/rpc/interceptors.go
```diff
@@ -0,0 +1,44 @@
+package rpc
+
+import (
+	"context"
+
+	"connectrpc.com/connect"
+)
+
+const SelectorMetadataKey = "flagd-selector"
```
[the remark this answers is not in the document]
toddbaert: Moved to an internal constant.

suthar26 (approved):

providers/flagd/internal/header/header.go
```diff
@@ -0,0 +1,5 @@
+// Package header defines metadata header keys shared by the flagd resolvers.
+package header
+
+// Selector is the metadata header used to pass the flag-set selector to flagd.
+const Selector = "flagd-selector"
```
JamieSinn: I don't love this being a named package; especially when it overlaps with the existing HTTP header name. Maybe something more like `metadataHeaders` or something more explicit?

toddbaert: renamed to `flagdmeta`.

JamieSinn (approved):

PR #902, opened 2026-06-19
feat(flipt): add environment-aware flag evaluation

### LINKED ISSUE #6053: Support Environments in openfeature provider
Hey folks!

It would be great if we could have support for Environments when using the `openfeature` provider. https://docs.flipt.io/v2/integration/openfeature#providers says that there's work in progress to support that, just curious if that's still the case at the moment

erka:
## This PR

- Add EnvironmentKey to EvaluationRequest and Environment field to Config
- Rename Evaluate to Variant, remove GetFlag from Service interface
- Update SDK auth: ClientTokenProvider → ClientAuthenticationProvider
- Default Environment to "default" with ForEnvironment option
- Update dependencies and regenerate mocks

closes flipt-io/flipt#6053


markphelps (approved):
nice one! thank you

providers/flipt/pkg/provider/provider.go
```diff
@@ -67,11 +67,11 @@ func WithService(svc Service) Option {
 	}
 }
 
-// WithClientTokenProvider sets the token provider for auth to support client
+// WithClientAuthProvider sets the client authentication provider for auth to support client
 // auth needs.
-func WithClientTokenProvider(tokenProvider sdk.ClientTokenProvider) Option {
```
toddbaert, line 72: This is breaking I think? Even though this is sub v1, we can mark it breaking with `!`. Release please will note that in the release notes, but only bump the minor according to sub v1 conventions due to [this config](https://github.com/open-feature/go-sdk-contrib/blob/d09a6ab662a6910811f7ba623c0a5ae3578287d9/release-please-config.json#L58).

providers/flipt/pkg/provider/provider.go
```diff
@@ -108,8 +116,8 @@ func NewProvider(opts ...Option) *Provider {
 			transport.WithHTTPClient(p.config.httpClient),
 			transport.WithCertificatePath(p.config.CertificatePath),
 		}
-		if p.config.TokenProvider != nil {
```
toddbaert, line 111: Same as https://github.com/open-feature/go-sdk-contrib/pull/902/changes#r3560043422

providers/flipt/go.mod
```diff
@@ -27,10 +30,20 @@ require (
...
 	google.golang.org/genproto/googleapis/api v0.0.0-20251202230838-ff82c1b0f217 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20251202230838-ff82c1b0f217 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 )
+
+// pin to v2.10.0
+replace (
+	go.flipt.io/flipt/rpc/flipt => go.flipt.io/flipt/rpc/flipt v1.54.1-0.20260612235747-ead29a1b4eda
+	go.flipt.io/flipt/sdk/go => go.flipt.io/flipt/sdk/go v0.13.1-0.20260608181651-f5bf5e9a38b5
+)
```
toddbaert: Slightly confused why this is needed? Seems like a no-op?

toddbaert (approved):
Approving but consider adding `!` (strong recommend from me).

erka:
Thanks for the review @toddbaert! I've updated the PR to avoid any breaking changes to the provider.

beeme1mr (approved):

PR #903, opened 2026-06-24
fix(flagd): resolve data race in InProcess service

oxddr:
This PR fixes a data race in the `InProcess` service of the `flagd` provider. The data race occurred during concurrent handling of provider error events and data synchronization updates.

## Root Cause
The `InProcess` service previously relied on `sync.Once` to manage state transitions (likely for signaling readiness). To allow signaling readiness multiple times (e.g., after a reconnection), the `sync.Once` instance was unsafely re-assigned (reset). This re-assignment is not thread-safe and caused data races when accessed concurrently by the event monitor and data sync process.

## Solution
1. **Replaced `sync.Once` with Mutex and Bool**: Replaced the state-tracking `sync.Once` with a `sync.Mutex` (`readyMu`) and a boolean flag (`ready`) in the `InProcess` struct.
2. **Safe Channel Sends**: Updated `processSyncData` and event handlers to modify the `ready` flag under the lock, but perform the channel sends (`i.events <- ...`) *outside* the lock to avoid potential deadlocks.
3. **Added Regression Test**: Added `TestInProcessServiceDataRace` in `service_test.go` which spawns concurrent goroutines sending error events and data updates to trigger the data race and verify the fix.

### DIFF
--- a/providers/flagd/pkg/service/in_process/grpc_sync.go
+++ b/providers/flagd/pkg/service/in_process/grpc_sync.go
@@ -207,7 +207,7 @@ func (g *Sync) Sync(ctx context.Context, dataSync chan<- sync.DataSync) error {
 					}
 				}
 			}
-			g.sendEvent(ctx, SyncEvent{event: of.ProviderError})
+			g.sendEvent(ctx, SyncEvent{Event: of.ProviderError})
 
 			if ctx.Err() != nil {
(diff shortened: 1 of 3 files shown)


toddbaert (approved):
Nice find! Thanks! I pushed one small lint fix.

PR #904, opened 2026-06-25
feat(codereadiness): add hook for version-based flag validation

marcin11858:
feat(codereadiness): Introduce the `codereadiness` hook to control feature flag evaluation by comparing the application's current version with a required minimum version specified in the flag's metadata. If the comparator returns an error the hook also returns an error to trigger fallback to the default flag value.

## This PR
Introduces the `codereadiness` hook under `hooks/codereadiness` to control feature flag evaluation based on the application's running code version.

### How it works
- Compares the application's current version with the required minimum version specified in the flag's metadata (by default, in the `minCodeVersion` key).
- If the comparator returns an error (by default, if the current version in SemVer format is lower than the minimum), the hook propagates the error to trigger the SDK's fallback mechanism, returning the flag's default value.
- The hook supports custom comparison logic (via `WithComparator`), custom metadata keys (via `WithMetadataMinVerKey`) and optional strict validation (via `WithValidationRequired`).

### Notes
- Includes comprehensive unit tests in `codereadiness_test.go`.
- Includes a `README.md` with setup and configuration examples.

### Follow-up Tasks
None.

### How to test
Run the package unit tests:
```bash
cd hooks/codereadiness
go test -v ./...

### CODE REVIEW

File: hooks/codereadiness/pkg/codereadiness.go [hunk 1]
```diff
@@ -0,0 +1,116 @@
...
+	of.UnimplementedHook
+	currentVersion     string
+	comparator         func(current, required string) error
+	validationRequired bool
+	metadataMinVerKey  string
+}
+
+func WithComparator(comparator func(current, required string) error) Option {
+	if comparator == nil {
+		panic("codereadiness: comparator cannot be nil")
```
oxddr: Do not panic. I'd move the validation to `New` function instead and check the resulting Hook struct. 

See: https://google.github.io/styleguide/go/decisions#dont-panic

hooks/codereadiness/codereadiness.go
```diff
@@ -0,0 +1,116 @@
...
+	return h
+}
+
+func (h Hook) After(ctx context.Context, hookContext of.HookContext, flagEvaluationDetails of.InterfaceEvaluationDetails, hookHints of.HookHints) error {
+	metadata := flagEvaluationDetails.FlagMetadata
+	if metadata == nil {
+		if h.validationRequired {
+			return fmt.Errorf("missing metadata error")
+		}
+		return nil
```
oxddr, line 105: I'd add logging to this hook and log in places like that, where you return early. In case, the user opt-ins to non-strict behaviour, this gives insight into why the validation did or did not kick in.

hooks/codereadiness/pkg/codereadiness.go
```diff
@@ -0,0 +1,116 @@
...
+		if h.validationRequired {
+			return fmt.Errorf("missing metadata error")
+		}
+		return nil
+	}
+
+	minCodeVersionInterface, ok := metadata[h.metadataMinVerKey]
+	if !ok {
+		if h.validationRequired {
+			return fmt.Errorf("missing %q error", h.metadataMinVerKey)
```
oxddr: nit: I'd try to be more descriptive with the error, something like `key %q missing in flag's %q metadata`

hooks/codereadiness/pkg/codereadiness.go
```diff
@@ -0,0 +1,116 @@
...
+	versionPrefix            = "v"
+)
+
+type Option func(*Hook)
+
+type Hook struct {
+	of.UnimplementedHook
+	currentVersion     string
+	comparator         func(current, required string) error
+	validationRequired bool
```
oxddr: personal, optional: I find name of this variable a bit misleading, especially when this is set to `false`. What about emphasising the **strict** behavior it enables, e.g `strictValidation`?

hooks/codereadiness/codereadiness.go
```diff
@@ -0,0 +1,116 @@
...
+			return fmt.Errorf("metadata %q is empty", h.metadataMinVerKey)
+		}
+		return nil
+	}
+	if h.currentVersion == "" {
+		if h.validationRequired {
+			return fmt.Errorf("missing current version error")
+		}
+		return nil
+	}
```
oxddr, line 130: I'd move this validation to `New` method, where `currentVersion` is provided, so that we fail at the startup, when this hook is likely created, not when the evaluation happens.

hooks/codereadiness/pkg/codereadiness.go
```diff
@@ -0,0 +1,116 @@
...
+	metadataMinVerKey  string
+}
+
+func WithComparator(comparator func(current, required string) error) Option {
```
oxddr: Please document all exported type and functions.

https://google.github.io/styleguide/go/decisions#doc-comments

hooks/codereadiness/pkg/codereadiness_test.go
```diff
@@ -0,0 +1,337 @@
...
+
+import (
+	"errors"
+	"strings"
+	"testing"
+
+	of "github.com/open-feature/go-sdk/openfeature"
+)
+
+func TestHook_After(t *testing.T) {
```
oxddr: Please use either table-drive tests (if tests share testing logic) or separate each test into its own function (if all tests are independent). Currently, it seems those subtests are fully independent and don't take any advantages of being run as subtests. In this case, having each be inside their own tests is more straightforward (which roughly falls under https://google.github.io/styleguide/go/guide#simplicity).

oxddr (commented):
This looks good overall, I left couple of comments regarding style and clarity of the code.

hooks/codereadiness/pkg/codereadiness.go
marcin11858: Done, moved the validation to `New` function

hooks/codereadiness/codereadiness.go
marcin11858: Done, added additional logs

hooks/codereadiness/pkg/codereadiness.go
marcin11858: Done, messages changed

hooks/codereadiness/pkg/codereadiness.go
marcin11858: Yes, it might be a bit misleading, changed to `strictValidation`

hooks/codereadiness/codereadiness.go
marcin11858: Validation moved

hooks/codereadiness/pkg/codereadiness.go
marcin11858: Comments added

hooks/codereadiness/pkg/codereadiness_test.go
marcin11858: Changed to independent testing

hooks/codereadiness/pkg/codereadiness.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	if minCodeVersion == "" {
+		if h.strictValidation {
+			return fmt.Errorf("metadata %q value is empty for flag %q", h.metadataMinVerKey, hookContext.FlagKey())
+		}
+		h.logger.DebugContext(ctx, "flag metadata key has an empty value, skipping validation", "metadataKey", h.metadataMinVerKey, "flagKey", hookContext.FlagKey())
+		return nil
+	}
+	if err := h.comparator(h.currentVersion, minCodeVersion); err != nil {
+		return fmt.Errorf("current version: %q required minimum version: %q flag %q, check failed: %w", h.currentVersion, minCodeVersion, hookContext.FlagKey(), err)
```
oxddr: Currently you're treating the same way validation (semver.IsValid) and the actual. I'd somewhat distinguish those errors. Invalid semver is a configuration error that should be fixed, failing validation is hook working as intended.

In case the validation fails, I'd also log it here.

hooks/codereadiness/pkg/codereadiness_test.go
```diff
@@ -0,0 +1,374 @@
+package codereadiness
+
+import (
+	"context"
+	"errors"
+	"log/slog"
+	"strings"
+	"testing"
+
+	of "github.com/open-feature/go-sdk/openfeature"
```
oxddr: nit: Please do not rename packages unless there is a conflict. 

https://google.github.io/styleguide/go/decisions#import-renaming

oxddr (approved):

hooks/codereadiness/codereadiness.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func defaultComparator(current, required string) error {
+	if !strings.HasPrefix(current, versionPrefix) {
+		current = versionPrefix + current
+	}
+	if !strings.HasPrefix(required, versionPrefix) {
+		required = versionPrefix + required
+	}
+	if !semver.IsValid(current) {
```
erka: Is there a way to do this only once for the current version, including lines 124–126? If the current version is invalid, perhaps the comparator should be implemented as a different function.

erka (commented):
Could we make this a flat package without `pkg` subpackage? I think we don't need it.

hooks/codereadiness/pkg/codereadiness_test.go
marcin11858: Done

hooks/codereadiness/pkg/codereadiness.go
marcin11858: I've added additional return variable which describe result of the comparison. Errors will stay for an unexpected situations e.g. wrong version format

marcin11858:
> Could we make this a flat package without `pkg` subpackage? I think we don't need it.

Done, removed `pkg` subpackage

hooks/codereadiness/codereadiness.go
marcin11858: Unfortunately I cannot be done once. SemVer format doesn't require `v` prefix but it is required by https://pkg.go.dev/golang.org/x/mod/semver library.

For better readability I've added `addRequiredSemVerPrefix` which adds required `v` prefix to SemVer.

.release-please-manifest.json
```diff
@@ -1,6 +1,7 @@
 {
     "hooks/open-telemetry": "0.3.7",
     "hooks/validator": "0.1.7",
+    "hooks/codereadiness": "0.1.0",
```
erka, line 4: just fyi - this will probably be released as 0.1.1 by release-please action

hooks/codereadiness/codereadiness.go
erka: If there is no custom comparator we could do more in `New` to prepare the `required` and validate it against semver.
If we know that it's invalid we could create and error and set comparator to 

```go
func New(...){
        //...
        required = addRequiredSemVerPrefix(required)
        if !semver.IsValid(required) {
		comparatorErr := fmt.Errorf("invalid required semver: %q", required)
		h.comparator = func(...) (bool, error) {return false, comparatorErr}
        }
        //...
}
// less work on hot path
func defaultComparator(current, required string) (bool, error) {
	current = addRequiredSemVerPrefix(current)
	if !semver.IsValid(current) { 
		return false, fmt.Errorf("invalid current semver: %q", current)
	}
	return semver.Compare(current, required) >= 0, nil
}
```

hooks/codereadiness/codereadiness.go
```diff
@@ -0,0 +1,148 @@
...
+	versionPrefix            = "v"
+)
+
+// Option is a functional option for configuring a Hook.
+type Option func(*Hook)
+
+// Hook validates min code version for After flag resolution.
+type Hook struct {
+	openfeature.UnimplementedHook
+	currentVersion    string
```
erka, line 34: could we rename this into `requiredVersion`?

marcin11858: This variable describe current version of the application in the binary. 

Do you mean about replace `minCodeVersion` by  `requiredVersion`?
I've choose `minCodeVersion` instead `requiredVersion` because it suggest that this is minimum version which works for a new feature (`>=`). `requiredVersion` can suggest that this version should exactly match current version (`==`). In addition if any software engineer will decide that any of feature should be enable only for selected version comparator (method: `WithComparator`) and metadata key (method: `WithMetadataMinVerKey`) can be override.

.release-please-manifest.json
marcin11858: Thank you for explanation

hooks/codereadiness/codereadiness.go
marcin11858: I agree, but in this case possibility of using custom comparator will be losing

hooks/codereadiness/codereadiness.go
erka: Oh... I misread it. The flag requires the minimal application version. Thank you for your explanation.

hooks/codereadiness/codereadiness.go
erka:
```diff
diff --git a/hooks/codereadiness/codereadiness.go b/hooks/codereadiness/codereadiness.go
index 260a161..7ee856d 100644
--- a/hooks/codereadiness/codereadiness.go
+++ b/hooks/codereadiness/codereadiness.go
@@ -25,6 +25,7 @@ type Hook struct {
 	openfeature.UnimplementedHook
 	currentVersion    string
 	comparator        func(current, required string) (bool, error)
+	customComparator  bool
 	strictValidation  bool
 	metadataMinVerKey string
 	logger            *slog.Logger
@@ -33,6 +34,7 @@ type Hook struct {
 // WithComparator sets a custom comparison function used to compare the current application version against the required minimum version.
 func WithComparator(comparator func(current, required string) (bool, error)) Option {
 	return func(h *Hook) {
+		h.customComparator = true
 		h.comparator = comparator
 	}
 }
@@ -78,6 +80,16 @@ func New(currentVersion string, opts ...Option) (*Hook, error) {
 	if h.currentVersion == "" {
 		return nil, errors.New("codereadiness: currentVersion cannot be empty")
 	}
+	if !h.customComparator {
+		h.currentVersion = addRequiredSemVerPrefix(currentVersion)
+		if !semver.IsValid(h.currentVersion) {
+			err := fmt.Errorf("invalid current semver: %q", currentVersion)
+			//return nil, err - may be better
+			h.comparator = func(current, required string) (bool, error) {
+				return false, err
+			}
+		}
+	}
 	return h, nil
 }
 
@@ -128,12 +140,10 @@ func (h *Hook) After(ctx context.Context, hookContext openfeature.HookContext, f
 	return nil
 }
 
+// defaultComparator reports whether the current semantic version satisfies
+// the flag required semantic version.
 func defaultComparator(current, required string) (bool, error) {
-	current = addRequiredSemVerPrefix(current)
 	required = addRequiredSemVerPrefix(required)
-	if !semver.IsValid(current) {
-		return false, fmt.Errorf("invalid current semver: %q", current)
-	}
 	if !semver.IsValid(required) {
 		return false, fmt.Errorf("invalid required semver: %q", required)
 	}
```
@marcin11858 wdyt?

erka (approved):
It looks great!

Just one small nit: we've started adding GoDoc examples, and it would be great to include one here as well. Something along the lines of the example in the OpenFeature Go SDK:

https://github.com/open-feature/go-sdk/blob/main/openfeature/isolated/isolated_example_test.go

toddbaert (dismissed):

toddbaert:
Code looks good ... one non-code question:

- What is the main use case for this hook in real-world scenarios? If app-version `v2.3.0` has my new feature, why would `v2.2.0` of my code have the flag for it? In other words, can you give some examples of when I would have a flag in a version `n-1` for a feature that should only be enabled in `n`? Is it for partially completed features? Like a code-level safety insurance?

marcin11858:
> Code looks good ... one non-code question:
> 
> * What is the main use case for this hook in real-world scenarios? If app-version `v2.3.0` has my new feature, why would `v2.2.0` of my code have the flag for it? In other words, can you give some examples of when I would have a flag in a version `n-1` for a feature that should only be enabled in `n`? Is it for partially completed features? Like a code-level safety insurance?

Thank you for your question. I found two possible use cases:

- working on a complex task: The developer introduced reading feature flag in one of the PR but there is more PRs in a queue which are waiting for a review. When all PRs will be merge the code is ready and new feature can be enable safety. 
- bug in existing application: App-version with a new feature (`v2.3.0`) was developed but it has a bug. To be sure that any instance of your service will not turn on the feature in app version `v2.3.0` the `minCodeVersion` can be changed to a new version with a fix e.g. `v2.3.1`. The developer is sure that is impossible to enable new feature in the `v2.3.0` app version on any instance.

marcin11858:
> It looks great!
> 
> Just one small nit: we've started adding GoDoc examples, and it would be great to include one here as well. Something along the lines of the example in the OpenFeature Go SDK:
> 
> https://github.com/open-feature/go-sdk/blob/main/openfeature/isolated/isolated_example_test.go

Done. Added some examples in the `hooks/codereadiness/example_test.go` file.

hooks/codereadiness/codereadiness.go
marcin11858: Another option might be creating an interface with two methods: 

```
// VersionComparator defines the interface for version comparison logic.
type VersionComparator interface {
	// Initialize is called once during Hook creation with the application's current version.
	Initialize(current string) error
	// Compare checks if the current version meets or exceeds the required version.
	Compare(required string) (bool, error)
}
```

Do you think should we split validation current version from `Compare` method?

marcin11858: Done. Split validation in two parts: `current` version during creation comparator `required` version in `After` method.

oxddr:
> > Code looks good ... one non-code question:
> > 
> > * What is the main use case for this hook in real-world scenarios? If app-version `v2.3.0` has my new feature, why would `v2.2.0` of my code have the flag for it? In other words, can you give some examples of when I would have a flag in a version `n-1` for a feature that should only be enabled in `n`? Is it for partially completed features? Like a code-level safety insurance?
> 
> Thank you for your question. I found two possible use cases:
> 
> * working on a complex task: The developer introduced reading feature flag in one of the PR but there is more PRs in a queue which are waiting for a review. When all PRs will be merge the code is ready and new feature can be enable safety.
> * bug in existing application: App-version with a new feature (`v2.3.0`) was developed but it has a bug. To be sure that any instance of your service will not turn on the feature in app version `v2.3.0` the `minCodeVersion` can be changed to a new version with a fix e.g. `v2.3.1`. The developer is sure that is impossible to enable new feature in the `v2.3.0` app version on any instance.

The usefulness of this check is much more apparent with trunk-based development, i.e. single, stable main branch, from which releases are cut on regular basis (e.g. daily or weekly). In this setup, it's common, that a partially completed feature ends up in the new release. This hook is to provide code-level safety - as pointed by you. Internally, we pair it with some other safety mechanisms. This is our last line of defence.

hooks/codereadiness/example_test.go
```diff
@@ -0,0 +1,120 @@
...
+	}
+
+	client := openfeature.NewClient("example-app")
+
+	// Since current version (v1.2.0) >= required (v1.0.0), flag evaluation succeeds.
+	enabled, err := client.BooleanValue(ctx, "feature-a", false, openfeature.EvaluationContext{})
+	if err != nil {
+		fmt.Println("Error evaluating flag:", err)
+		return
+	}
```
erka:
```suggestion
	client := openfeature.NewDefaultClient()

	// Since current version (v1.2.0) >= required (v1.0.0), flag evaluation succeeds.
	enabled := client.Boolean(ctx, "feature-a", false, openfeature.EvaluationContext{})
```
some nit for the example

hooks/codereadiness/example_test.go
```diff
@@ -0,0 +1,120 @@
...
+
+	// Because strict validation is enabled and the expected metadata key is missing,
+	// evaluation fails and returns the default value (false).
+	enabled, err := client.BooleanValue(ctx, "feature-b", false, openfeature.EvaluationContext{})
+	fmt.Println("feature-b enabled:", enabled)
+	fmt.Println("evaluation error:", err != nil)
+
+	// Output:
+	// feature-b enabled: false
+	// evaluation error: true
```
erka:
```suggestion
	client := openfeature.NewDefaultClient()

	// Because strict validation is enabled and the expected metadata key is missing,
	// evaluation fails and returns the default value (false).
	enabled, err := client.BooleanValue(ctx, "feature-b", false, openfeature.EvaluationContext{})
	fmt.Println("feature-b enabled:", enabled)
	fmt.Println("evaluation error:", err)

	// Output:
	// feature-b enabled: false
	// evaluation error: after hook: key "requiredCodeVersion" missing in flag's "feature-b" metadata
```

marcin11858: Done

hooks/codereadiness/example_test.go
marcin11858: Done

toddbaert:
I'm ready to merge this, but please consider my question [here](https://github.com/open-feature/spec/pull/405#issuecomment-4984772676) first! :pray:

toddbaert (approved):

erka (approved):

beeme1mr (approved):

PR #920, opened 2026-07-10
chore: do not transitively update go version with deps

toddbaert:
Update CI and Makefile to fail if we update Go version via transitive dep.

@thomaspoignant I think this is coming from GoFF:

```
providers/go-feature-flag-in-process  (go 1.26.4)
        |
        +-- github.com/thomaspoignant/go-feature-flag v1.54.1  (go 1.26.4)
```


thomaspoignant:
@toddbaert I am not sure what to do here, GOFF last version needs a higher version of GO than the contrib repo supports. I am not sure what to do here 🤷‍♂️

toddbaert:
> @toddbaert I am not sure what to do here, GOFF last version needs a higher version of GO than the contrib repo supports. I am not sure what to do here 🤷‍♂️

Of course it's no problem for GoFF to require 1.26 - that's totally up to you as the maintainer; it's even fine for your provider to require the same, but it's a problem for the contribs workspace; I certainly want to keep flagd for instance supporting 1.25, which is in-line with OpenFeature's policy of supporting actively maintained runtimes (both Go 1.25/1.26) in this case.

Would you be OK with if the GoFF OpenFeature provider hosted here lags a bit?

Seems like Go 1.25 support will end in August, so I don't think it will generally be very far behind... We could even release it and then revert, so we'd have a version compatible with 1.26 out there for a while. That's up to you.

erka (approved):

JamieSinn (approved):

thomaspoignant:
Yes this is fine.

toddbaert:
> Yes this is fine.

OK - reverting here. I opened https://github.com/open-feature/go-sdk-contrib/issues/924 as a reminder.

PR #927, opened 2026-07-26
fix(from-env): safely compare complex criteria values

### LINKED ISSUE #759: from-env provider: potential panic when comparing evalCtx values
high

While changing `interface{}` to `any` is a great modernization step, there's a potential pre-existing issue in this function's implementation. The comparison `val != criteria.Value` on line 37 can cause a panic if the values from the evaluation context or the flag definition are complex types like slices or maps, which are not comparable in Go. Using `reflect.DeepEqual` for this comparison would be safer and prevent potential panics. For example: `if !ok || !reflect.DeepEqual(val, criteria.Value)`. Since the function body is not part of this diff, I'm noting this here for future improvement.

_Originally posted by @gemini-code-assist[bot] in https://github.com/open-feature/go-sdk-contrib/pull/754#discussion_r2443530303_

Baltsat:
## Summary

- compare targeting criteria with `reflect.DeepEqual`
- prevent evaluation panics for slice and map attributes
- cover both complex value shapes through the public provider API

Fixes #759

## Validation

- `go test ./... -count=1`
- `go test -race ./... -count=1`
- `go vet ./...`
- `golangci-lint v2.8.0 run ./...`
- production-only revert reproduces the `comparing uncomparable type []interface {}` panic; restoring the fix passes

The commands above were run in `providers/from-env` with Go 1.25.12. A repository-wide module sweep also passed outside `providers/flagd/e2e`; that package requires the absent `flagd-testbed/gherkin/config.feature` fixture.

## AI-assisted contribution

AI assistance was used for investigation, implementation, testing, and review. I reviewed the final diff and verified all reported commands locally.


erka (changes requested):
Hey @Baltsat. Thanks for your PR!

While this doesn't panic, it introduces unnecessary overhead during evaluations. Since comparable values can be checked directly, could you use `reflect.DeepEqual` only for non-comparable values?

Baltsat:
Thanks — updated in 7709677. Comparable values with the same dynamic type now use direct `==`; only same-type non-comparable values fall back to `reflect.DeepEqual`. I also added a distinct-pointer regression that fails under unconditional deep equality, while retaining the public slice/map regression.

The isolated focused package, race, vet, formatting, and production-revert checks pass. The branch is mergeable and DCO is green; repository checks are now running/pending their normal workflow gate.

providers/from-env/pkg/eval.go
```diff
@@ -21,6 +23,17 @@ type Criteria struct {
...
+func valuesEqual(left, right any) bool {
+	leftType := reflect.TypeOf(left)
+	if leftType != reflect.TypeOf(right) {
+		return false
+	}
+	if leftType == nil || leftType.Comparable() {
+		return left == right
+	}
+	return reflect.DeepEqual(left, right)
+}
```
erka, line 72:
```suggestion
// valuesEqual reports whether left and right represent equal values.
//
// It is intended for comparing decoded JSON-like data (the result of
// json.Unmarshal into any), but falls back to reflection so it also
// behaves correctly for arbitrary comparable and non-comparable types.
//
// Common JSON types (string, float64, bool, nil, []any, map[string]any)
// are compared directly via a type switch, avoiding reflection.
// For all other types, left and right must share the same concrete type
// (as reported by reflect.TypeOf) to be considered equal; values of
// differing concrete types — e.g. int32(5) and int64(5) — are always
// unequal, even if numerically equivalent. Comparable types (including
// typed nil pointers) are then compared with ==. Non-comparable types
// (structs or arrays containing slices/maps/funcs, etc.) fall back to
// reflect.DeepEqual.
func valuesEqual(left, right any) bool {
	switch l := left.(type) {
	case nil:
		return right == nil
	case string:
		r, ok := right.(string)
		return ok && l == r
	case float64:
		r, ok := right.(float64)
		return ok && l == r
	case bool:
		r, ok := right.(bool)
		return ok && l == r
	case []any:
		r, ok := right.([]any)
		return ok && slices.EqualFunc(l, r, valuesEqual)
	case map[string]any:
		r, ok := right.(map[string]any)
		return ok && maps.EqualFunc(l, r, valuesEqual)
	}
	leftType := reflect.TypeOf(left)
	if leftType != reflect.TypeOf(right) {
		return false
	}
	if leftType.Comparable() {
		return left == right
	}
	return reflect.DeepEqual(left, right)
}

```

erka (changes requested):
Please move the tests from `providers/from-env/pkg/noncomparable_criteria_test.go` into `providers/from-env/pkg/eval_test.go`, then remove `noncomparable_criteria_test.go`.

If possible, it would also be great to add a few more test cases to improve coverage while you're touching these tests.

Baltsat:
Addressed the requested test organization and coverage update:

- moved the slice/map regressions into `providers/from-env/pkg/eval_test.go`
- removed `providers/from-env/pkg/noncomparable_criteria_test.go`
- kept the matrix focused on equal/unequal slices, equal/unequal maps, equal/unequal comparable values, distinct pointers, and mismatched dynamic types
- left `providers/from-env/pkg/eval.go` unchanged in this follow-up

Validated with Go 1.25.8 from `providers/from-env`:

- `gofmt -d pkg/eval.go pkg/eval_test.go`
- `go test ./pkg -count=1`
- `go test -race ./pkg -count=1`
- `go vet ./pkg`
- `go test ./... -count=1`
- `go vet ./...`
- `git diff --check`

The offline module proxy was limited to the four existing `go.sum`-pinned modules; `go mod verify` reported `all modules verified`. A production-only revert reproduced `panic: runtime error: comparing uncomparable type []interface {}` for the equal-slice case, and the restored final tree passed package and race tests.

This follow-up was implemented and validated with OpenAI Codex assistance.

erka (approved):
Thank you @Baltsat for your work!

Baltsat:
Thanks, @erka. I plan to keep contributing across go-sdk-contrib and flagd rather than make this a one-off; I have also landed open-feature/flagd#2006. I can take responsibility for reproducing and triaging provider issues and doing evidence-backed reviews. I am also working toward OpenFeature organization membership. Once this PR merges, would you be open to considering sponsorship, or tell me what additional contribution you would want to see first? I know the process requires prior agreement from two approvers or maintainers from different companies.

erka:
I don't think I'm the right person to answer questions about organization membership. Those decisions are made by other organization members.

sahidvelji (approved):

sahidvelji:
@Baltsat Thanks for the contribution. I suggest joining the #openfeature Slack channel in the CNCF Slack. You can find the link at the bottom of the OpenFeature site https://openfeature.dev/

PR #933, opened 2026-08-11
fix(flagsmith): handle float64 values in FloatEvaluation

16ajbm:
## This PR

In `providers/flagsmith/pkg/provider.go`, `FloatEvaluation` returns `TYPE_MISMATCH` when a flag value is a native JSON number (e.g. `"feature_state_value": 10`). Go's `encoding/json` decodes JSON numbers as `float64`, but `FloatEvaluation` only handles `string` values. `IntEvaluation` in the same file already handles `float64` correctly, with a comment referencing https://pkg.go.dev/encoding/json#Unmarshal as reasoning. This PR applies the same pattern to `FloatEvaluation`: check for `float64` first, then fall back to `string` parsing.

- adds `float64` type assertion before the existing `string` fallback in FloatEvaluation
- adds `numeric_flag` test fixture (`feature_state_value: 100.1`, a native JSON number)
- adds test case: evaluating a native JSON number (`numeric_flag`) through FloatEvaluation

### Related Issues

No related issues.

Discovered while building a vendor-neutral OF harness that maps a single "number" flag type to `FloatValueDetails` (since `float64` is Go's representation of a JSON number). `IntEvaluation` works for integer values, but consumers using `FloatValueDetails` as a generic numeric evaluation path cannot resolve flags whose values are native JSON numbers.

### Notes

The Flagsmith API returns integer values as native JSON numbers (e.g. `10`, not `"10"`). `encoding/json` decodes these as `float64`. `FloatEvaluation` assumes values are always strings (comment states "Because We store floats as string"), which holds for decimal values set via the dashboard but not for integer values set via the API.

### How to test

```bash
cd providers/flagsmith && go test ./pkg/ -run TestFloatEvaluation -v
```


erka:
Hey @16ajbm.

Could you point me to any evidence that there’s an issue with the current implementation? The comment in the code says, `// Because We store floats as string`, and I also created a float flag and tested it myself - it seems to work correctly.

I suspect the issue might be that the input you’re using doesn’t match the actual data returned by the server. Could you share an example of the real data that’s causing the problem? That would help us verify what’s happening.

toddbaert:
> Hey @16ajbm.
> 
> Could you point me to any evidence that there’s an issue with the current implementation? The comment in the code says, `// Because We store floats as string`, and I also created a float flag and tested it myself - it seems to work correctly.
> 
> I suspect the issue might be that the input you’re using doesn’t match the actual data returned by the server. Could you share an example of the real data that’s causing the problem? That would help us verify what’s happening.

cc @matthewelwell

16ajbm:
Hey @erka, 

Thank you for getting back to me! Thanks @toddbaert for cc'ing @matthewelwell. This change is proposed as I'm looking into integrating with Flagsmith, among other providers, for Avid Technology. We are using a comparison test harness to test the compatibility with different providers. 

You're correct in that the input doesn't exactly match the data returned by the server. We have an integer flag created via API, like so:

```sh
curl -X POST \
  -H "Authorization: Api-Key <token>" \
  -H "Content-Type: application/json" \
  -d '{"feature":{"name":"max-items"},"environment_default":{"enabled":true,"value":{"type":"integer","value":"10"}}}' \
  "https://api.flagsmith.com/api/experiments/environments/<env_key>/update-flag-v2/"
```

When the client acquires this flag it gets `"feature_state_value": 10` which is a native JSON number, it is unmarshalled as `float64` - as the `IntEvaluation` [accounts for](https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagsmith/pkg/provider.go#L265). All good here.

Where it differs/the reason for this proposed change is that we are trying to stay vendor neutral. If we have the restriction of a single numeric type in our harness (matching JSON's single numeric type), we would need to utilize `FloatEvaluation` for both integers and floats. This works in the other providers within the repo, for example [Unleash](https://github.com/open-feature/go-sdk-contrib/blob/main/providers/unleash/pkg/provider.go#L110). This does not work for the Flagsmith provider.

This PR was intended as a point of discussion. This change would be backwards compatible, leaving the direct evaluation on float strings from the API untouched. It would also allow us to call `FloatEvaluation` directly, not having to consider the different flag types/couple to the provider. 

Thanks again for taking the time to look at this, happy to discuss further.

erka:
@16ajbm Harness is really cool, and I admire the work you’re doing in this area. We’d love to learn more about the challenges you’re facing here.

That said, numbers themselves are another puzzle, and different providers are trying to solve it in different ways. Let’s see what @matthewelwell thinks about your approach and how it could impact the implementation in other languages.

emyller:
While @matthewelwell is out today, and acknowledging our data model decision, it may be useful to bring up supporting references:

- [Kotlin provider](https://github.com/Flagsmith/flagsmith-openfeature-provider-kotlin/blob/ca9813e04c723214f1b23239b01025bb385af8d0/provider/src/main/java/com/flagsmith/openfeature/FlagsmithProvider.kt#L118-L123)
- [PHP provider](https://github.com/open-feature/php-sdk-contrib/blob/04efa3e91b0451994611d805f601149fdf13b094/providers/Flagsmith/src/service/FlagEvaluator.php#L505-L518)
- [NodeJS](https://github.com/open-feature/js-sdk-contrib/blob/878213ffbb656fdb802f6be5740e0c17a75b294f/libs/providers/flagsmith/src/lib/type-factory.ts#L5-L9)

providers/flagsmith/pkg/provider_test.go
```diff
@@ -164,7 +164,22 @@ const FlagsJson = `
...
+	{
+    "id": 6,
+    "feature": {
+        "id": 6,
+        "name": "numeric_flag",
+        "default_enabled": true,
+        "type": "STANDARD",
+        "project": 1
+    },
+    "feature_state_value": 100.1,
```
matthewelwell: If I understand correctly, this isn't actually the use case that we're testing for, since the Flagsmith API will return a float value as a string. 

Should we change this to correctly represent the specific scenario that we care about? Maybe with a comment explaining why (or, perhaps even better, it's own dedicated test to describe the functionality)? 

```suggestion
    "feature_state_value": 100,
```

matthewelwell (commented):
Thanks @16ajbm for the PR, and for all the details in the discussion - I think I'm caught up now. If I understand correctly, this really just boils down to wanting to ensure that `parseFloatValue` returns the actual value for any numeric value returned from the API, instead of returning a `TYPE_MISMATCH` error. 

This sounds reasonable to me, and looks as though how it is already handled in other Flagsmith providers, as per @emyller's comment above. 

I'd be inclined to approve this PR, taking my one specific comment into account.

providers/flagsmith/pkg/provider_test.go
16ajbm: Made the suggested change in the test fixtures, as well as the relevant assertion in the test case. I've added a comment and explicit value in the test case, as follows:

```
    // The Flagsmith API returns integer values as native JSON numbers,
    // which encoding/json decodes as float64.
    name:                "Should resolve a numeric JSON value as float",
    flagKey:             "numeric_flag",
    expectedValue:       float64(100),
```
the expected value could be put under a variable like `flagKey: expectedNumericFlagValue` or something if preferred, but only one instance so I though explicit might be fine and readable

erka:
@16ajbm could you please sign your commits to resolve DCO [issue](https://github.com/open-feature/go-sdk-contrib/pull/933/checks?check_run_id=94841025847)

16ajbm:
@matthewelwell I see you were assigned recently, commits are signed and PR is ready to go at your earliest convenience

matthewelwell (approved):

providers/flagsmith/pkg/provider_test.go
matthewelwell: Thanks - looks good.

matthewelwell:
Thanks @16ajbm, all looks good to me from a Flagsmith perspective. We'll just need one of the @open-feature/sdk-golang-maintainers to approve as well.

erka (approved):

toddbaert (approved):

PR #949, opened 2026-09-03
feat(flagd)!: add missing options, use gRPC on wire

### LINKED ISSUE #936: flagd: align retry defaults with spec, fix retry `maxAttempts`, implement missing config options
We must implement open-feature/flagd#2027, which proposes changing the spec defaults so the stream-reconnect backoff can't leave us disconnected longer than the stale grace period.

- `pkg/service/in_process/grpc_config.go:17` `DefaultRetryBackoffMaxMs`: 12000 -> 5000
- `pkg/configuration.go:35` `defaultGracePeriod`: 5 -> 10

`pkg/service/in_process/grpc_config.go:55` also sets `MaxAttempts` to 3, but the spec's retry policy specifies 4 (the initial attempt plus retries at 1s, 2s, 4s); should be 3 -> 4. Harmless to fix now since nothing clamps at a 5000 cap.

We're also missing three options the spec lists, most importantly `streamDeadlineMs` (`FLAGD_STREAM_DEADLINE_MS`, default 600000, rpc and in-process), which the spec describes as an application-layer keepalive. Without it we never recycle an idle stream, so we're the only provider with no keepalive on the sync stream. Also absent: `keepAliveTime` (`FLAGD_KEEP_ALIVE_TIME_MS`) and `offlinePollMs` (`FLAGD_OFFLINE_POLL_MS`).

These should be implemented together: with 5000/10 a deadline-driven recycle reconnects well inside the grace period, yielding a STALE/READY pair and no ERROR. Blocked on flagd#2027.

leakonvalinka:
:warning: this PR switches from using the Connect wire proto for `evaluation.v2` to gRPC for `evaluation.v2`; not a behavioral change, but possible impacts on infra such as proxy rules.

# Conformance fixes in accordance to spec

## Added:
- streamDeadlineMs - FLAGD_STREAM_DEADLINE_MS
- keepAliveTime - FLAGD_KEEP_ALIVE_TIME_MS
- offlinePollMs - FLAGD_OFFLINE_POLL_MS

## Updated:
- retryGracePeriod 5 -> 10
- retryBackoffMaxMs 12000 -> 5000
- gRPC retry MaxAttempts 3 -> 4

## Related:
- un-skipped variant e2e test suites related to the above (see [rpc_test.go](https://github.com/open-feature/go-sdk-contrib/pull/949/changes#diff-403d719101b4b7c18882f5001a01766e2696c44c3d191c4a17ada2a281501c62))


toddbaert:
@leakonvalinka looks like the CI is green now - is this good to review?

leakonvalinka:
@toddbaert Sorry, I didn't get around to removing the work-around in the grpc service until now - now the CI reflects the failing e2e tests (due to the clash of the retry policy, 8s outage and 10s waiting assertion). You'll have to download the full logs file tho because it's so long that Github cut it off at some point. Unfortunately I wasn't able to find a quick fix :/

providers/flagd/e2e/file_test.go
```diff
@@ -41,7 +41,7 @@ func TestFileProviderE2E(t *testing.T) {
 
 	// Run tests with file-specific tags, focusing on core evaluation scenarios
 	// Skip complex connection-related and synchronization scenarios for file provider
-	tags := "@file && ~@reconnect && ~@sync && ~@grace && ~@events && ~@unixsocket && ~@metadata && ~@deprecated && ~@fractional-v1"
+	tags := "@file && ~@reconnect && ~@sync && ~@grace && ~@events && ~@unixsocket && ~@metadata && ~@deprecated && ~@fractional-v1 && ~@fractional-v3"
```
toddbaert, line 44: Unrelated new test not yet implemented.

providers/flagd/pkg/service/in_process/grpc_config.go
```diff
@@ -52,7 +52,7 @@ func (g *Sync) buildRetryPolicy() string {
 					{"service": "flagd.sync.v1.FlagSyncService"},
 				},
 				"retryPolicy": RetryPolicy{
-					MaxAttempts:          3,
+					MaxAttempts:          4,
```
toddbaert, line 55: Thanks @leakonvalinka  - yep, this is all spec compliant now.

providers/flagd/e2e/rpc_test.go
```diff
@@ -25,8 +25,14 @@ func TestRPCProviderE2E(t *testing.T) {
...
-	// Run tests with RPC-specific tags - exclude unimplemented scenarios
-	tags := "@rpc && ~@unixsocket && ~@targetURI && ~@sync && ~@metadata && ~@grace && ~@customCert && ~@caching && ~@forbidden && ~@deprecated && ~@fractional-v1"
+	// Run tests with RPC-specific tags - exclude unimplemented or inapplicable scenarios
+	tags := "@rpc" + // rpc resolver scenarios
+		" && ~@unixsocket" + // unix socket channel not supported
+		" && ~@targetURI" + // target-uri scenarios not supported
+		" && ~@customCert" + // custom TLS cert not wired up
+		" && ~@deprecated" +
+		" && ~@fractional-v1" + // legacy fractional algorithm
+		" && ~@fractional-v3" // cbor fractional not yet implemented
```
toddbaert, line 35: No longer skipped:

- `@grace`
- `@forbidden`
- `@metadata`
- `@caching`

All these are implemented and working now.

providers/flagd/pkg/service/rpc/service.go
```diff
@@ -713,6 +803,8 @@ func newClient(cfg Configuration) (schemaConnectV2.ServiceClient, error) {
 
 	// build options
 	var options []connect.ClientOption
+	// speak the gRPC wire protocol (like the other flagd providers) for compatibility with gRPC proxies
+	options = append(options, connect.WithGRPC())
```
toddbaert, line 829: This is a "breaking change" in terms of wire behavior - but doesn't change behavior otherwise, and this makes it consistent with every other impl.

toddbaert (approved):
@leakonvalinka I pushed a few additions namely:

- implemented the grace period in RPC mode
- switched to gRPC wire protocol which allowed us to add some additional test suites
- un-skipped a few test-suites that were already working (caching, for example).

providers/flagd/pkg/service/rpc/service.go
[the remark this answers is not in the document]
toddbaert: I'll have to come back to this.

providers/flagd/pkg/service/in_process/service_file_test.go
```diff
@@ -51,3 +51,55 @@ func TestInProcessOfflineMode(t *testing.T) {
...
+	}
+
+	service := NewInProcessService(Configuration{OfflineFlagSource: offlinePath, OfflinePollMs: 100})
+
+	if err := service.Init(); err != nil {
+		t.Fatal(err)
+	}
+	defer service.Shutdown()
+
+	if detail := service.ResolveBoolean(context.Background(), "myBoolFlag", false, make(map[string]interface{})); !detail.Value {
```
erka:
```suggestion
	if detail := service.ResolveBoolean(t.Context(), "myBoolFlag", false, make(map[string]any{})); !detail.Value {
```

providers/flagd/pkg/service/in_process/service_file_test.go
```diff
@@ -51,3 +51,55 @@ func TestInProcessOfflineMode(t *testing.T) {
...
+		}
+	}`
+	if err := os.WriteFile(offlinePath, []byte(updated), 0644); err != nil {
+		t.Fatal(err)
+	}
+
+	// then - the change is detected within a few poll intervals
+	deadline := time.Now().Add(2 * time.Second)
+	for {
+		detail := service.ResolveBoolean(context.Background(), "myBoolFlag", true, make(map[string]interface{}))
```
erka:
```suggestion
		detail := service.ResolveBoolean(t.Context(), "myBoolFlag", true, make(map[string]any{}))
```

providers/flagd/pkg/service/in_process/grpc_sync.go
```diff
@@ -246,7 +260,16 @@ func (g *Sync) performSyncCycle(ctx context.Context, dataSync chan<- sync.DataSy
...
 
 	// Handle the stream with proper context cancellation
-	return g.handleFlagSync(ctx, stream, dataSync)
+	err = g.handleFlagSync(streamCtx, stream, dataSync)
+
+	// If the configured stream deadline elapsed (and the parent context is still live), this is an
+	// intentional stream recycle - signal a clean cycle so the caller reconnects without error/backoff.
+	if g.StreamDeadlineMs > 0 && errors.Is(streamCtx.Err(), context.DeadlineExceeded) && ctx.Err() == nil {
+		g.Logger.Debug("stream deadline reached, recycling sync stream")
+		return nil
```
erka, line 269: no backoff in `Sync`?

providers/flagd/pkg/service/rpc/service.go
```diff
@@ -91,7 +108,9 @@ func (s *Service) Init() error {
 	}
 
 	ctx, cancelFunc := context.WithCancel(context.Background())
+	s.cancelMu.Lock()
 	s.cancelHook = cancelFunc
```
erka, line 117: I have a suspicion that `Service.Init` could be called multiple times if `Provider.Init` encounters an issue and doesn't follow happy path with `initialized`.  Maybe it would be good to check whether it holds something and cancel it, similar to `Shutdown`,

toddbaert: You are right, and the changes in this PR to make it block longer to make Connect work more like gRPC (connect doesn't have the [waitForReady](https://grpc.io/docs/guides/wait-for-ready/) concept) make it more likely to happen.

I extracted some of the shutdown logic into a new func that we can call to make sure nothing is leaked.

providers/flagd/pkg/service/in_process/grpc_sync.go
toddbaert: There's 2 things here:

- gRPC's built-in L4 connection backoff + waitForReady semantics, and method-level retry policy
- our `retryBackoffMaxMs` which sets a hard delay during the retry loop, even for errors not in the retry policy (this IS basically an app-level backoff in the outer loop here)

Just for background:

These settings both come from real world experience. Over the years we've seen some really bad bugs related to proxies and gRPC where proxies (especially host-local ones like envoy) return errors immediately if they can't reach the upstream, causing tight loops and CPU spikes as the provider tries to connect thousands of times a minute. Both of these in combination give robustness and backstop these loops.

tests/flagd/testframework/provider_steps.go
```diff
@@ -46,7 +46,23 @@ func (s *TestState) createProviderInstance() error {
...
+		Option:    "RetryBackoffMaxMs",
+		ValueType: "Integer",
+		Value:     "500",
+	})
+	s.ProviderOptions = append(s.ProviderOptions, ProviderOption{
+		// generous init deadline so a slow first connect under load doesn't fail readiness
+		Option:    "deadlineMs",
+		ValueType: "Integer",
+		Value:     "15000",
 	})
```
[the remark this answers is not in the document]
toddbaert: This should be resolved now.

providers/flagd/pkg/service/in_process/grpc_sync.go
erka: I'm looking into this loop, and it seems that if `nil` is returned the backoff logic isn't applied and `performSyncCycle` is called again.

https://github.com/open-feature/go-sdk-contrib/blob/main/providers/flagd/pkg/service/in_process/grpc_sync.go#L178-L226

providers/flagd/pkg/service/in_process/grpc_sync.go
```diff
@@ -42,6 +43,8 @@ type Sync struct {
 	RetryGracePeriod        int
 	RetryBackOffMs          int
 	RetryBackOffMaxMs       int
+	StreamDeadlineMs        int
+	KeepAliveTime           int64
```
JamieSinn, line 47:
```suggestion
	KeepAliveTimeMs           int64
```

For naming consistency - since we multiply it by MS lower down.

providers/flagd/pkg/service/in_process/service.go
```diff
@@ -119,6 +120,8 @@ type Configuration struct {
 	RetryBackOffMaxMs       int
 	FatalStatusCodes        []string
 	DeadlineMs              int
+	StreamDeadlineMs        int
+	KeepAliveTime           int64
```
JamieSinn, line 124: Same `Ms` suffix comment here

providers/flagd/pkg/service/in_process/service_file_test.go
```diff
@@ -51,3 +51,55 @@ func TestInProcessOfflineMode(t *testing.T) {
...
+		  "myBoolFlag": {
+			"state": "ENABLED",
+			"variants": {
+			  "on": true,
+			  "off": false
+			},
+			"defaultVariant": "off"
+		  }
+		}
+	}`
```
JamieSinn, line 88: part of me wants this as the native go struct and then to serialize it out. 

another part of me is perfectly fine with this formatting. 

nit mostly.

providers/flagd/pkg/service/in_process/service.go
toddbaert: This is dictated by the spec, and asserted.

providers/flagd/pkg/service/in_process/grpc_sync.go
toddbaert: This is dictated by the spec, and asserted.

providers/flagd/pkg/service/in_process/service.go
JamieSinn: Small eye twitch moment there lol

JamieSinn (approved):
nothing other than the nits from me

providers/flagd/pkg/service/in_process/grpc_sync.go
toddbaert: Yes, but `nil` is only returned on stream deadline (which is many minutes for streams, 10m in our case by default). This isn't a tight-loop situation and helps us to re-build the stream quickly.

toddbaert: The background here: gRPC keepalive frames don't transit proxies, so stream deadline acts sort of a s "application level" keep alive to make sure the stream is fresh and hasn't dropped between proxies.

erka (approved):

PR #956, opened 2026-09-10
feat(provider-tck): let adopters add their own scenarios and steps

aepfli:
Stacked on #944. Answers @toddbaert's review request on open-feature/spec#423: an adopter with provider-specific behaviour must be able to run their own scenarios inside the TCK's backend lifecycle rather than in a parallel harness.

## What an adopter writes

```go
tck.Run(t, tck.Config{
    // ... Name, NewProvider, Control, Capabilities as before ...
    ExtensionFeatures: os.DirFS("testdata/tck-extensions"),
    ExtensionSteps: func(ctx *godog.ScenarioContext) {
        ctx.Step(`^the fractional bucket for "([^"]*)" is "([^"]*)"$`, theBucketIs)
    },
})
```

`ExtensionFeatures` is any `fs.FS`; every `.feature` file in it joins the run. `ExtensionSteps` is called during scenario initialisation after the TCK's own registrations, so an extension step sees the same context and hooks, and reaches the provider under test through `tck.ClientFromContext(ctx)` — the provider is registered under a suite-scoped domain the adopter never names. Extension features can and usually should use the canonical steps: `Given a stable provider` in a `Background` is what puts them in the same lifecycle phase as the canonical scenarios.

Both fields are optional. A `Config` without them hands godog the same `fs.FS` and the same path as before.

## Why configuration rather than convention

Java scans the classpath for glue packages and Python picks up a `conftest.py`, so both get zero-configuration extension. Go has no runtime scanning — a step definition is a function that has to be called — so this is configuration. The constraint is that it stays two fields and no test infrastructure of the adopter's own.

## Shadowing

In Java a same-named file in a second classpath root silently *replaced* the canonical one and the suite went green having run the adopter's version. Here the canonical assets keep their `assets/gherkin/` paths and extensions are mounted under `extensions/`, so nothing an adopter supplies is reachable under the canonical prefix. `testdata/tck-shadow/evaluation.feature` copies the canonical file name, feature name and a scenario name; the test asserts the canonical file was still parsed byte for byte and all 29 canonical scenarios ran unchanged.

The two sources are mounted as one filesystem rather than passed through two godog mechanisms. `Options.FeatureContents` looks like the shortcut, but godog parses it and `Options.Paths` in separate calls with separately seeded id generators, so AST node, pickle and pickle-step ids collide across the two sets — and this package keys skip reasons and Messages test cases by pickle id.

## Second commit, separable

Nothing stopped a run reporting success having executed fewer canonical scenarios. Verified: `go test -run 'TestEnvelopeReferencesTheResults/An_unknown_flag_key_returns_the_code_default'` passed green and wrote a well-formed report covering 1 of 29 scenarios. The second commit compares what executed against the embedded assets compiled by godog's own parser and fails the run on any canonical scenario that produced no outcome. A capability-gated scenario is not a gap — it ran the gate and is in the results as `SKIPPED` with its reason. Extension scenarios are counted and reported but can never close a gap.

Drop this commit if you would rather it went separately; nothing in the first commit depends on it.

## Note on the pending options refactor

@erka asked on #940 for `tck.Run` to take functional options instead of one `Config` struct, and for the package to move to `tools/tck` without `pkg/`. Struct fields here are deliberate so the two changes do not collide — that refactor subsumes these into `WithFeatures(...)`/`WithSteps(...)` and should land on its own.

## Verification

- `go build`, `go vet`, `gofmt` clean; package tests pass with `-count=1`.
- One suite runs 29 canonical and 1 extension scenario against one backend lifecycle, and the extension step resolves through the TCK's client.
- Reports produced by the commit before this branch and by its tip are identical apart from the payload digest: same envelope, same 29 scenarios, same feature sources, same step outcome tallies.
- Envelopes still validate against `conformance-report.schema.json` (Draft 2020-12), with and without extensions configured.


aepfli:
Folded into the chain: the extension point is in the base PR (#940, commit `ca75476`) and the canonical-set guard is in the report PR (#944). Nothing on this branch is absent from those two — every file it carried exists on one of them — so it has nothing left of its own. Closing to keep the chain a four-branch progression: base → report, base → flagd → ofrep.

